PR #7, opened 2016-02-22
Add ntp tests

jkonecny12:
They are not tested. I didn't have time to create boot iso and I don't know how to use updates image with the kickstart test.

Is there possible way how to test it with updates image?

*Related: rhbz#1309396*

### DIFF
--- a/ntp-nontp-without-chrony-gui.ks.in
+++ b/ntp-nontp-without-chrony-gui.ks.in
@@ -1,0 +1,36 @@
+#version=DEVEL
+url @KSTEST_URL@
+install
+network --bootproto=dhcp
+
+# test timezone in GUI mode
+# with --nontp and -chrony
+timezone --utc --nontp Europe/Prague
+
+bootloader --timeout=1
+zerombr
+clearpart --all --initlabel
+autopart
+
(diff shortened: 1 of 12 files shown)


clumens:
Yeah, PRs that add or modify tests will be run through jenkins automatically.  The only thing is that to avoid spamming the test system, I have it skip anything that touches more than three tests.  It looks like yours does.  If you split it up a little bit it should work, or I can raise the number a bit.

clumens:
I removed that check because it doesn't seem relevant right now.  However I can't find a way to mark this PR as "changed" so our jenkins job will pick it up and re-run the check on it.  Could you do something stupid to a file (like add a comment or a blank line or something) somewhere to change it?  That should force the test to re-run.  Then we'll get useful data.

jkonecny12:
OK I will do the change but the code I want to test is here https://github.com/rhinstaller/anaconda/pull/517  now. So these test are going to fail now.

When the patch will be in master I will do the force push here.

jkonecny12:
Updated now

jkonecny12:
Finally it passed. I think it should be OK for merging now.

PR #11, opened 2016-04-04
use meshlab not RepetierHost as 3d-printing canary

AdamWill:
RepetierHost has been 'optional' since F23, so it's not a good
canary for this test any more.


bcl:
Looks ok to me.

PR #14, opened 2016-04-06
rework proxy test generation, make tests more robust

AdamWill:
While working on running kickstart-tests in openQA we noticed
that the proxy tests had some unspoken assumptions: that the
repo URL would be a direct mirror URL (not a mirror list) and
that it would use HTTP (not HTTPS).

While fixing that, I noticed there's quite a bit of duplication
between proxy-auth and proxy-kickstart, and proxy-cmdline is
more or less a subset of them only they had diverged quite a
lot in style. I also thought the mechanism for getting the proxy
code into the kickstart %pre - which requires both make *and*
ksflatten - seemed a bit over-complex for really just sticking
chunks of text together.

This is my suggestion for resolving most of the problems. It uses
a Makefile to produce all three proxy .ks.in files from the
same template plus a few snippet files and proxy.py, and makes
the tests more robust against use of mirror lists and HTTPS
for the base repo. Caveats:

* It does not cover use of HTTPS for the addon repo, that is
  still required to be served as HTTP. We *could* allow for
  this by just weakening a couple of the proxy-repo.inc tests
  for the 'httpslist' and 'httpsdir' cases, I'm not sure if
  it's worth it.

* proxy.py seems not to be capable of proxying HTTPS traffic
  with password authentication enabled, so you still can't
  do proxy-auth with an HTTPS mirror or mirrorlist. This is
  a pre-existing issue, I believe.

Note I *have not tested this* with the official runner yet, only with openQA and direct manual testing in a VM. I think the changes to the runner and the `.sh` files are correct, but this could stand testing.


AdamWill:
@dashea - @clumens and I chatted a bit about this on IRC:

<clumens> adamw: i will tell you that one thing i've consciously been trying to do with kickstart-tests is not be too clever with reducing duplication.  i'd rather be able to look at something and know what it's testing than have to piece it together.
<adamw> clumens: well, the proxy tests are already pulling in the proxy.py with a Makefile and ksflatten.
<clumens> yeah i do kinda hate it.
<adamw> and there's clearly been some kinda divergence between cmdline and the other two.
<clumens> i'm not saying "NO!!!", but just that i don't like making things more complicated
<adamw> sure
<adamw> i understand
<adamw> it's very easy to run 'make' and get a complete .ks.in with only the standard subs, though
<adamw> (including the proxy code)
<clumens> without really digging into it, i'm inclined to merge it if it fixes something
<adamw> i'm fine with it being dug into honestly =)
<adamw> we can see what davidshea thinks too?
<clumens> yeah he's more familiar with that
<adamw> if you'd like i could do a pr to simplify the proxy.py inclusion, a pr to sync cmdline with the other two, and a pr to pull the 'work with https / mirror list' changes into all three separate .ks.ins, i guess

So, basically, we're kinda looking for your opinion on this. I'm willing to do a more conservative approach as mentioned in the last comment, but only if you'd definitely prefer that, I do kinda not like the duplication. So if you could let us know whether you like this, the approach where we simplify the `proxy.py` inclusion and fix for HTTPS / mirror lists but leave the `.ks.in` file duplication, or if you have another idea, that'd be great! thanks.

dashea:
I'm not a fan of having two ways to generate the actual .ks stuff. And it is a lot harder to tell what's going on in the source files for the proxy tests now, so based on that I'm leaning more toward just leaving some level of duplication in.

What do you mean by simplifying the proxy.py inclusion? The change in the PR is doing basically the same thing, just in a different place and three times instead of once.

AdamWill:
"I'm not a fan of having two ways to generate the actual .ks stuff."

Not quite sure what you mean there. This isn't an extra way to generate the `.ks` files, it's a different way to handle the additional work required for the proxy `.ks.in` files before the standard `.ks.in`-to-`.ks` procedure actually works. They're already outliers in this sense because they're the only files where you have to do something special before producing a `.ks` will actually work (that is, create `proxy-common.ks`). The only other case where `Makefile.prereqs` is used is a bit different, because the kickstart generation works fine in that case whether you've produced `install.img` or not, `install.img` is just an extra test asset.

"What do you mean by simplifying the proxy.py inclusion? The change in the PR is doing basically the same thing, just in a different place and three times instead of once."

No, it isn't; the current code uses `make` to produce `proxy-common.ks` out of `proxy.py` by just adding a few lines to it, then uses `ksflatten` to (effectively) stick `proxy-common.ks` into each proxy file. Two different tools and mechanisms just to stick some boilerplate into a few files. My version does all the work in the same place (the `Makefile`).

By 'simplify' I mean 'just do it with one tool', i.e. get rid of the use of ksflatten and just do it with a substitution, whether that's driven by `make` or just done in `run-kickstart-tests` or whatever. If we keep the duplicated `.ks.in` files that would mean including the `%pre` boilerplate in those then just having a placeholder text where `proxy.py` would be replaced in.

clumens:
I believe we solved this satisfactorily with PR # 16.  If that's not the case, feel free to reopen and we'll take another look.  Thanks!

AdamWill:
well, I still kinda liked this more, but since you didn't, #16 was my more stripped-down offering. :)

PR #15, opened 2016-04-07
reqpart, default-fstype: ensure just one result from blkid

AdamWill:
running reqpart in openQA, I found blkid would actually return
three results for this search, so the output would be 'xfs\n
xfs\nxfs' (not ext4 because I'm testing with a Server image).
The -l option protects against this by returning only one
result for the search, the one with the 'highest priority' (per
the man page).

### DIFF
--- a/default-fstype.ks.in
+++ b/default-fstype.ks.in
@@ -22,12 +22,12 @@ shutdown
 %end
 
 %post
-root_fstype=$(blkid -o value -t LABEL=root -s TYPE)
+root_fstype=$(blkid -o value -t LABEL=root -s TYPE -l)
 if [ $root_fstype != "ext4" ]; then
     echo "default fstype is incorrect (got $root_fstype; expected ext4)" >> /root/RESULT
 fi
 
-boot_fstype=$(blkid -o value -t LABEL=boot -s TYPE)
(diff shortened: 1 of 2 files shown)


AdamWill:
Just for the record, in case anyone wanted to dig into this and figure out why something (I guess?) changed, my reproducer is to install with [F24 Alpha Server netinst](https://dl.fedoraproject.org/pub/alt/stage/24_Alpha-1.7/Server/x86_64/iso/Fedora-Server-netinst-x86_64-24_Alpha-1.7.iso) and `inst.ks=https://www.happyassassin.net/ks/reqpart-bug.ks` (which is just `reqpart.ks` with no shutdown or reboot after install). Wait for install to complete, and do `chroot /mnt/sysimage; blkid -t LABEL=root`. For me, that results in:

    /dev/mapper/fedora-root: LABEL="root" UUID="5f4af8b3-014a-41f1-8b6e-db68ba25a7a0" TYPE="xfs"
    /dev/dm-2: LABEL="root" UUID="5f4af8b3-014a-41f1-8b6e-db68ba25a7a0" TYPE="xfs"
    /dev/block/253:2: LABEL="root" UUID="5f4af8b3-014a-41f1-8b6e-db68ba25a7a0" TYPE="xfs"

so you can see it finds the same device (UUID is the same) three different ways.

PR #17, opened 2016-04-14
Start NFS for KSTEST_NFS_ADDON_REPO if not specified

atodorov:


dashea:
I like the idea of making the nfs tests more self-contained, but I'm really uncomfortable with the idea of messing with system NFS configs from a test script. Maybe there's a userspace NFS server that could be run from kickstart-tests?

clumens:
Yeah, I don't think it's the test case's job to change the configuration of the system.  If there's not a way to run some other NFS server that is more standalone, I don't really want to pursue doing this.

atodorov:
I'm closing this PR. I've managed to move the code internally and prepare the environment before the test is executed.

PR #18, opened 2016-04-14
Skip libvirt IPs

atodorov:
The script is taking the first IP it finds, which in my case is 192.168.122.1 and stuff doesn't appear to work for some reason. I have a br0 configured which shows at the end of the ip addr list output.

### DIFF
--- a/scripts/find-ip
+++ b/scripts/find-ip
@@ -21,6 +21,7 @@
 # Print something that looks like an externally accessible IPv4 address.
 
 ip addr list | \
+    grep -v '192.168.122.1' | \
     grep '^[[:space:]]*inet\>.*scope global' | \
     head -1 | \
     sed 's/^[[:space:]]*inet \([[:digit:].]*\).*/\1/'


dashea:
First, this is specific to the particular libvirt config, right? Is 192.168.122 even the default? Mine uses 192.168.200.* for the kickstart-tests, though I don't know if that's something I messed with or not.

Second, don't we want to use a libvirt address? This is for communication between the hypervisor running httpd and a guest running the kickstart-test, so it seems like if the script can find the right IP for that, that would be ideal. Maybe there's a way we could query libvirt for the address to use?

bcl:
I'm not sure how reliable this is, but you can get the MAC assigned to a vm using 'sudo virsh domiflist', parse the output, and then grep /proc/net/arp for the MAC to get the IP address.

jkonecny12:
No response on this PR for more than 2 years so I'm closing it now.
Also we have plans to rewrite structure of these tests to make them more friendly for use in other than virt environment.

PR #22, opened 2016-06-03
Add a test for different cases of encrypted swap

dashea:
This will fail on finding the /etc/fstab entry for the swap on RAID (but everything before that ought to work). The RAID case will be fixed by the second commit of https://github.com/rhinstaller/anaconda/pull/659.

### DIFF
--- a/encrypt-swap.ks.in
+++ b/encrypt-swap.ks.in
@@ -1,0 +1,93 @@
+#version=DEVEL
+url @KSTEST_URL@
+install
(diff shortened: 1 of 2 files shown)


dashea:
Missed a "!" in the very first test, trying again.

vpodzime:
Looks good to me.

dashea:
All the pieces on the anaconda side should be in place now, so: jenkins, test this please

PR #31, opened 2016-08-23
Http server is now using ports from 40000 to 41000

jkonecny12:
It's nice to use specific ports to open all ports on server.

### DIFF
--- a/scripts/httpd.py
+++ b/scripts/httpd.py
@@ -32,8 +32,18 @@ if len(sys.argv) != 2:
     print("Usage: httpd.py <directory>", file=sys.stderr)
     sys.exit(1)
 
+server=None
 # Bind to any free port
-server = HTTPServer(('', 0), SimpleHTTPRequestHandler)
+for port in range(40000, 41000):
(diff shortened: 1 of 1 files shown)


scripts/httpd.py
```diff
@@ -32,8 +32,17 @@
...
 
+server=None
 # Bind to any free port
-server = HTTPServer(('', 0), SimpleHTTPRequestHandler)
+for port in range(40000, 41000):
+    try:
+        server = HTTPServer(('', port), SimpleHTTPRequestHandler)
+    except OSError:
+        continue
+    break
```
vpodzime: I think this would be better readable if the ``break`` was moved into an ``else`` branch of the try-except block.

vpodzime:
Looks good to me otherwise.

scripts/httpd.py
jkonecny12: Nice improvement! I'll do an update.

jkonecny12:
Updated, thank you.

PR #32, opened 2016-08-24
Fix lvm-raid-1 test be reducing space requirement

jkonecny12:
There were 8G disks and the raid in kickstart file created partitions for 8000 size. Test was failing because the partitions can't be created.


vpodzime:
Looks good to me.

PR #34, opened 2016-08-24
Adapt to changes in LVM cache size handling

vpodzime:
Blivet now makes sure that the overall space required for the LVM cache(s)
equals the value passed as the requested cache size. Thus size of the metadata
space is subtracted from the cache's size and for the first cache defined the
same applies also to the size of the pmspare space.


jkonecny12:
Looks good to me. Thank you for your help.

PR #38, opened 2016-09-19
Fix copy out from kickstart tests

jkonecny12:
When one of the virt-copy-out files will fail nothing is copied out.

Changed to for cycle instead.


jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

PR #39, opened 2016-09-19
Add missing coverage test to proxy-cmdline KS test

jkonecny12:
Parameters `debug=1` and `inst.debug` are required to enable coverage tests.

---------------------------------
This test will fail because of bug which is fixed in PR https://github.com/rhinstaller/kickstart-tests/pull/38 .

### DIFF
--- a/proxy-cmdline.sh
+++ b/proxy-cmdline.sh
@@ -26,7 +26,7 @@ prereqs() {
 }
 
 kernel_args() {
-    echo vnc inst.proxy=http://127.0.0.1:8080
+    echo vnc inst.proxy=http://127.0.0.1:8080 inst.debug
 }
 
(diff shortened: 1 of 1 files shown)


vpodzime:
Are you sure with the ``debug=1`` option? I remember that causing quite a lot of troubles.

jkonecny12:
I'll test it, these are default arguments which are in all tests without  `kernel_args()` function.

jkonecny12:
It looks like the `debug=1` is not required so I removed it.

vpodzime:
Looks good to me.

jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

PR #44, opened 2016-10-21
network: consolidate some %post checks into functions

rvykydal:
I want to use them also in other network tests. We might start to think
about kickstart preprocessing step for including bash functions (or anything).


jkonecny12:
The preprocessing step could be nice. I'll create item in JIRA for this.

PR #48, opened 2016-11-10
Ignore the rsyslogd critical error

jkonecny12:
This error doesn't block the installation.

### DIFF
--- a/scripts/run_one_ks.sh
+++ b/scripts/run_one_ks.sh
@@ -17,6 +17,7 @@
 # Red Hat, Inc.
 #
 # Red Hat Author(s): Chris Lumens <clumens@redhat.com>
+#                    Jiri Konecny <jkonecny@redhat.com>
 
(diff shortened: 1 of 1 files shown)


scripts/run_one_ks.sh
```diff
@@ -111,37 +111,47 @@ runone() {
...
-    echo RESULT:${name}:SUCCESS
+    ret=$(validate ${tmpdir})
+    if [[ $? != 0 ]]; then
+        RESULT="FAILED:${ret}"
+        RET_CODE=1
+    fi
+
+    if [ -z "${RESULT}" ]; then
+        RESULT="SUCCESS"
+    fi
```
vpodzime: This all looks good to me, but I don't like the mixture of ``[`` and ``[[``. Please use only one or another (preferably ``[[``), but not both.

jkonecny12:
**UPDATED**

PR #49, opened 2016-11-11
Master network static configuration tests

rvykydal:


scripts/create-network.py
```diff
@@ -0,0 +1,81 @@
...
+
+# Look for free 192.168 subnet assuming 255.255.255.0 netmask.
+# As 'default' network has.
+
+used_subnets = set()
+for network in conn.listAllNetworks():
+    network_xml = network.XMLDesc()
+    root = ET.fromstring(network_xml)
+    for ip in root.iter('ip'):
+        used_subnets.add(ip.attrib['address'][:11])
```
jkonecny12: Maybe I'm wrong but I see problem here, that this is not atomic operation.
The tests are started on 4 VMs at once so when you read this settings from libvirt it could be before any of the other 3 VM will be started but when you use it a few lines later in the code then 3 of them could be running with their subnet but you are still using the old value.

I think we need some global synchronization mechanism or better to use some global lock.

rvykydal:
Good point, I've added file based locking and removed the mis-copy&pasted comment.

PR #61, opened 2017-01-04
Master add iscsi tests

rvykydal:


rvykydal:
I reworked the patch a bit:
- The iscsi.sh test now actually installs to iscsi target. Getting results from iscsi disk image was added to validate() function. 
- Autopartitioning is used instead of specifying disks by device name in custom partitioning which may be fragile.
- The disk image is removed based on -k run_kickstart_tests.sh option (for values < 2 which remove disk images in general).

iscsi.sh
```diff
@@ -0,0 +1,78 @@
...
+    ### Tear down iscsi target
+    local imgfile=${tmpdir}/${iscsi_disk_img}
+    local test_id=$(basename ${tmpdir})
+    # use only lower case in target name
+    local lc_test_id=${test_id,,}
+    local backstore=${lc_test_id}
+    local wwn=iqn.2003-01.kickstart.test:${lc_test_id}
+    local logfile=${tmpdir}/targetcli.log
+    remove_iscsi_target $wwn $backstore $imgfile $logfile
+
```
jkonecny12: You can remove this blank line.

vpodzime:
Very nice, thanks for this!

PR #64, opened 2017-01-17
Master fix team test sh sharing

rvykydal:


rvykydal:
Wrong branch, closing

rvykydal:
Or better, fixing the branch.

team.sh
```diff
@@ -24,7 +24,7 @@
 # injected in initrd) network devices are not found in sysfs in the time of
 # parsing the kickstart.
 
-TESTTYPE="network"
+TESTTYPE=${TESTTYPE:="network"}
```
jkonecny12: `:=` should be `:-` because now you are doing substitution twice.

```
${parameter:-word}
              Use Default Values.  If parameter is unset or null, the expansion of word is substituted.  Otherwise, the value of parameter is substituted.
${parameter:=word}
              Assign  Default Values.  If parameter is unset or null, the expansion of word is assigned to parameter.  The value of parameter is then substituted.  Positional parameters and special parame‐
              ters may not be assigned to in this way.
```

rvykydal: Hmm, good to know, thanks!

PR #66, opened 2017-01-17
Fix hostname test on RHEL system

jkonecny12:
Hostname test do not need `validate()` in the shell script, because it's using autopart.
This `validate()` worked only for Fedora system because of disk name prefix `fedora-*`.

### DIFF
--- a/hostname.sh
+++ b/hostname.sh
@@ -20,26 +20,3 @@
 TESTTYPE="network"
 
 . ${KSTESTDIR}/functions.sh
-
-validate() {
-    disksdir=$1
-    args=$(for d in ${disksdir}/disk-*img; do echo -a ${d}; done)
-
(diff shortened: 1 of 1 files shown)


rvykydal:
We (blivet) derive default names of volume groups from system (fedora) and hostname. I am not sure if we want to check it as a part of this test. But if we want, we should rather check it in %post IMO, so I am ok with not overriding the validate function.
Given we had bug reports like this https://bugzilla.redhat.com/show_bug.cgi?id=1359631 maybe creating a separate test for just this (hostname-defaultvgname.sh) could be useful.

jkonecny12:
Most tests solving this by giving custom label to the `/`. However I think we want to use as little as possible of a custom code in .sh files so IMHO this should be removed.

About the bug, it seems to me like a good idea to create tests for this 👍 .

PR #74, opened 2017-02-06
Add updates images support

jkonecny12:
``Usage: ./scripts/run_kickstart_tests.sh -i [boot.iso] -k [1|2] -u [link to updates.img] [test.sh...]``

Updates image must be on the reachable server in network for VM.

The `-u` option is then past to the inst.updates kernel arguments for the VM. This feature can break tests and give you non-valid results, but it is much quicker then building your own `boot.iso`.

### DIFF
--- a/scripts/run_kickstart_tests.sh
+++ b/scripts/run_kickstart_tests.sh
@@ -58,9 +58,13 @@ IMAGE="${TEST_BOOT_ISO}"
 # 2 - Keep log files and disk images (will take up a lot of space)
 KEEPIT=${KEEPIT:-0}
 
+# Link to a server where an updates image is stored. Use on your own
+# responsibility, this can break tests.
+UPDATES_IMG=""
+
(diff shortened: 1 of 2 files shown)


scripts/run_kickstart_tests.sh
```diff
@@ -79,9 +83,14 @@ while getopts ":i:k:t:" opt; do
            # more than one type.  We'll do a pretty stupid test for it.
            TESTTYPE=$OPTARG
            ;;
-
+       u)
+           # Link to an updates image on a server. This will be add as a kernel parameter
```
rvykydal: will be added

jkonecny12:
I guess it should work as you said.

scripts/run_kickstart_tests.sh
jkonecny12: I will fix that. Thanks!

PR #80, opened 2017-03-21
Add network tests for changing static ipv4 config to dhcp

rvykydal:
The tests test issues in rhbz#1433891 rhbz#1432886 which are reported for
change in UI but the root cause makes also change in %pre of kickstart fail.

### DIFF
--- a/network-static-to-dhcp-pre-single.sh
+++ b/network-static-to-dhcp-pre-single.sh
@@ -1,0 +1,74 @@
+#
+# Copyright (C) 2017  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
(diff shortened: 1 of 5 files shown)


rvykydal:
jenkins, test this please

network-static-to-dhcp-pre-single.sh
```diff
@@ -0,0 +1,74 @@
+#
+# Copyright (C) 2015  Red Hat, Inc.
```
jkonecny12: I guess it's 2017 now :)

network-static-to-dhcp-pre.sh
```diff
@@ -0,0 +1,75 @@
+#
+# Copyright (C) 2015  Red Hat, Inc.
```
jkonecny12: The same here.

PR #85, opened 2017-05-10
Add test for hostname defined via boot options.

rvykydal:


hostname-bootopts.sh
```diff
@@ -0,0 +1,73 @@
...
+    local network=$(basename ${tmpdir})
+
+    local scriptdir=${PWD}/scripts
+    local ips="$(${scriptdir}/create-network.py "${network}")"
+    local ip="$(echo "$ips" | cut -d ' ' -f 1)"
+    local netmask="$(echo "$ips" | cut -d ' ' -f 2)"
+    local gateway="$(echo "$ips" | cut -d ' ' -f 3)"
+    local hostname=testhostname.example.com
+
+    echo "ip_static_boot_config=ip=${ip}::${gateway}:${netmask}:${hostname}:ens4:none" > ${tmpdir}/ip_static_boot_config
```
jkonecny12: Is here some special reason which I'm not aware of to explain why do you need `${tmpdir}/ip_static_boot_config` file? Why don't you use the `ip_static_boot_config=""` variable above?

hostname-bootopts.sh
```diff
@@ -0,0 +1,73 @@
...
+    ### Create dedicated network to prevent IP address conflicts for parallel tests
+
+    local network=$(basename ${tmpdir})
+
+    local scriptdir=${PWD}/scripts
+    local ips="$(${scriptdir}/create-network.py "${network}")"
+    local ip="$(echo "$ips" | cut -d ' ' -f 1)"
+    local netmask="$(echo "$ips" | cut -d ' ' -f 2)"
+    local gateway="$(echo "$ips" | cut -d ' ' -f 3)"
+    local hostname=testhostname.example.com
```
jkonecny12: Isn't this generic? I mean the whole create network and store it's values which is generated here. Can't it be used for some other tests and if yes then move it to some other shared file please.

hostname-bootopts.sh
rvykydal: It wouldn't work.
I think the reason is that the prepare function is run in a subshell (command substitution).
In run_one_test.sh:
ksfile=$(prepare ${ks} ${tmpdir})

hostname-bootopts.sh
rvykydal: I don't think it would be worthy here. The function from shared file would just pass the return value of the script create-network.py and in a test calling the function we'd need to parse the return value into ip, netmask, gateway in a similar way.

jkonecny12: OK, we can move it later if there will be another use for that.

hostname-bootopts.sh
jkonecny12: Ahh damn shell.. :(

rvykydal:
Also removed from hostname-bootopts.sh (as jkonecny12 proposed face to face)
ip_static_boot_config=""

PR #92, opened 2017-05-31
network: add test for onboot policy

rvykydal:
In rhel we make sure at least one the devices activated during installations
has ONBOOT set to yes.

This test handles the case where ONBOOT was set to yes for additional device
because we were looking only on ONBOOT values in ksdata and a device activated
via boot options without ks configuration is not in ksdata for automatic
installations (ie if network spoke is not visited) so we were wrongly assuming
there is no device with ONBOOT=yes.

Note: the originally failing test for this issue was network-static, but
network-static is now replaced by more comprehensive network-static-2 tests and
this issue will be tested by this new test.

### DIFF
--- a/onboot-bootopts-pre.ks.in
+++ b/onboot-bootopts-pre.ks.in
@@ -1,7 +1,11 @@
-#test name: network-static
+#test name: onboot-bootopts-pre
 url @KSTEST_URL@
 install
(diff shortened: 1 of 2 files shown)


rvykydal:
Ah, this is already merged

PR #93, opened 2017-06-01
network: add test for legacy (rhel6) boot options

rvykydal:


rvykydal:
I tagged the test as knownfailure for now, see the commit message.

PR #101, opened 2017-08-23
network tests - fix MAC address conflict between tests

rvykydal:


rvykydal:
Added, though I don't think it is useful. To mitigate the risk of conflict in the future I added the info about statically assigned MACs (esp. how to list all used MACs) in https://github.com/rhinstaller/kickstart-tests/pull/102.
Anyways, thanks for the review.

PR #104, opened 2017-09-19
Master bindtomac

rvykydal:


post-lib-network.sh
```diff
@@ -46,6 +46,33 @@ function device_ifcfg_key_missing() {
     fi
 }
 
+function check_ifcfg_key_exists() {
```
jkonecny12: Some comment for the function describing arguments and behavior is really needed here.

jkonecny12: Truth is that is should be everywhere here. However, this should be done as separate PR I guess.

rvykydal:
I added the comments.
Also I noticed that there are some redundant functions in the lib so I removed them.

post-lib-network.sh
```diff
@@ -31,21 +35,39 @@ function check_device_connected() {
...
 
+# check_device_ifcfg_bound_to_mac NIC
+# Check that the ifcfg file of device NIC is bound to MAC address
+function check_device_ifcfg_bound_to_mac() {
+    local nic="$1"
+    check_ifcfg_key_exists $nic DEVICE no
+    check_ifcfg_key_exists $nic HWADDR yes
+}
+
+# check_bond_has_slave BOND SLAVE "yes|no"
```
jkonecny12: This should be "yes"|"no" from the point of consistency.

post-lib-network.sh
```diff
@@ -95,29 +123,8 @@ function check_number_of_device_ipv4_addresses() {
...
-        exit_code=1
-    fi
-
-    cat /proc/net/bonding/${bond} | egrep -q "^Slave.*${slave}"
-    if [[ $? -ne ${exit_code} ]]; then
-        echo "*** Failed check: ${bond} has slave ${slave} ${expected_result}" >> /root/RESULT
-    fi
-}
-
+# check_team_has_slave TEAM SLAVE "yes|no"
```
jkonecny12: The same here.

rvykydal:
The comments should be fixed now, thanks for the review.

PR #106, opened 2017-10-05
Test the install method hmc with a fake support for SE/HMC

poncovka:
This install method can be used only on s390x with SE/HMC. To be
able to test it, we use fake scripts that will mount DVD the usual
way.

This test requires a full DVD ISO, so it is marked as a known failure.
It can be used only for accessing the packages on the DVD. It cannot
be used to get stage2.

Related: rhbz#1289918

### DIFF
--- a/hmc.ks.in
+++ b/hmc.ks.in
@@ -1,0 +1,56 @@
+#version=DEVEL
+#test name: hmc
+
+# This test is for testing the install method hmc.
+#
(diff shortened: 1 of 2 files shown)


poncovka:
@jkonecny12 But the installation method has to be `hmc`, not `url`.

poncovka:
Some changes:
* the installation method `hmc` is selected in `hmc.ks.in`
* options for manual testing were moved to `hmc.sh`

PR #130, opened 2018-06-07
Master fix new python scripts

jkonecny12:
Many fixes to make the new python scripts usable.

### DIFF
--- a/scripts/launcher/lib/validator.py
+++ b/scripts/launcher/lib/validator.py
@@ -22,6 +22,10 @@
 
 import re
 
+from .test_logging import get_logger
+
+log = get_logger()
+
 
 def replace_new_lines(line):
         line.replace("#012", "\n")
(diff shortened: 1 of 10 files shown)


jkonecny12:
**UPDATED**:
* Fix another smaller issues.
* Change reporting.
* Fix missing return codes from shell functions.
* Fix run_report script

I need to also find a way how to add Traceback information to `run_report.sh` if we want that.

scripts/launcher/lib/validator.py
```diff
@@ -71,20 +72,14 @@ def return_code(self):
...
-                                                       self._result_msg,
-                                                       description)
-            if self._return_code != 0:
-                self._log.error(msg)
-            else:
-                self._log.info(msg)
+    def report_result(self, description=""):
+        msg = self._result_formatter.format_result(self.result,
+                                                   self._result_msg,
+                                                   description)
```
M4rtinK: The description parameter should be dropped here as well.

jkonecny12:
**UPDATED:**
* Fixed issue found by @M4rtinK

PR #132, opened 2018-06-12
Use rsync for remote copy & keep last used image around

M4rtinK:
Use rsync instead of scp to copy the kickstart-tests folder and boot iso
to the remote test runners and also keep the last used boot iso
around (we only keep the last used image and delete any others).

This way we will get a much faster test startup on remote runners as
only changes in the kickstart test will be transferred and the image
transfer skipped altogether if the same image is used.

For synchronizing the kickstart-tests folder the rsync --delete flag
if used, so no artifacts created during the tests runs in the folder
should remain once rsync is done with the sync.

### DIFF
--- a/scripts/run_kickstart_tests.sh
+++ b/scripts/run_kickstart_tests.sh
@@ -304,16 +304,21 @@ if [[ "$TEST_REMOTES" != "" ]]; then
     # parallel doesn't like globs, and we need to put the boot image somewhere
     # that qemu on the remote systems can read.
     for remote in ${TEST_REMOTES}; do
-        ssh kstest@${remote} mkdir kickstart-tests
-        scp -r . kstest@${remote}:kickstart-tests/
-        scp ${IMAGE} kstest@${remote}:kickstart-tests/
+        ssh kstest@${remote} mkdir -p kickstart-tests
+        ssh kstest@${remote} mkdir -p install_images
+        rsync -az --delete . kstest@${remote}:kickstart-tests/
+        rsync -az ${IMAGE} kstest@${remote}:install_images/
+        # remove any old images
+        ssh kstest@${remote} find install_images -type f ! -name ${_IMAGE} -delete
     done
 
     # (1a) We also need to copy the provided image to under kickstart_tests/ on
     # the local system too.  This is because parallel will attempt to run the
     # same command line on every system and that requires the image to also be
     # in the same location.
-    cp ${IMAGE} ${_IMAGE}
+    mkdir -p ../install_images
+    cp ${IMAGE} ../install_images/
+    find ../install_images -type f ! -name ${_IMAGE} -delete
 
     # (2) Run parallel.  By default add the local system to the list of machines
     # being passed to parallel.
@@ -329,7 +334,7 @@ if [[ "$TEST_REMOTES" != "" ]]; then
     cd ..
     parallel --no-notice ${remote_args} --wd kickstart-tests --jobs ${TEST_JOBS:-4} \
              sudo PYTHONPATH=$PYTHONPATH scripts/launcher/run_one_test.py \
-                                                               -i ${_IMAGE} \
+                                                               -i ../install_images/${_IMAGE} \
                                                                -k ${KEEPIT} \
                                                                ${UPDATES_ARG} ${BOOT_ARG} {} ::: ${tests}
     rc=$?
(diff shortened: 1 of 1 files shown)


scripts/run_kickstart_tests.sh
```diff
@@ -305,8 +305,13 @@ if [[ "$TEST_REMOTES" != "" ]]; then
...
         ssh kstest@${remote} mkdir kickstart-tests
-        scp -r . kstest@${remote}:kickstart-tests/
-        scp ${IMAGE} kstest@${remote}:kickstart-tests/
+        ssh kstest@${remote} mkdir install_images
+        rsync -az --delete . kstest@${remote}:kickstart-tests/
+        rsync -az ${IMAGE} kstest@${remote}:install_images/
+        # remove any old images
+        ssh kstest@${remote} find install_images -type f ! -name ${_IMAGE} -delete
+        # copy the image to the kickstart-tests folder, so guestfs can use it
+        ssh kstest@${remote} install_images/${_IMAGE} kickstart-tests/
```
jkonecny12: Maybe I'm blind but aren't you missing remote command here?

jkonecny12: Or maybe there should be `scp` instead of `ssh`?

scripts/run_kickstart_tests.sh
```diff
@@ -350,11 +355,8 @@ if [[ "$TEST_REMOTES" != "" ]]; then
             scp -r kstest@${remote}:/var/tmp/kstest-\* /var/tmp/
         fi
 
-        ssh kstest@${remote} sudo rm -rf kickstart-tests /var/tmp/kstest-\*
+        ssh kstest@${remote} sudo rm -rf /var/tmp/kstest-\*
     done
-
-    # (3a) And then also remove the copy of the image we made earlier.
-    rm ${_IMAGE}
```
jkonecny12: Why are you removing this?

M4rtinK: We are running rsync with  --delete on the kickstart-tests folder, so we don't need to explicitly delete the ISO file anymore - rsync will do that for us when it takes care of any leftovers in the folder on the next run.

scripts/run_kickstart_tests.sh
M4rtinK: That's a good point - that does not look correct. I wonder how it still all works like this. :D

M4rtinK:
Updated the PR:
- drop the attempt to copy the iso into the kickstart-tests folder as that's actually no longer needed
- use the install_images folder also locally when running with TEST_REMOTES_ONLY != yes

scripts/run_kickstart_tests.sh
jkonecny12: I'm not that sure, this is local system call `rm` but those rsync calls are for remote machines. Is that really working correctly?

scripts/run_kickstart_tests.sh
```diff
@@ -329,7 +334,7 @@ if [[ "$TEST_REMOTES" != "" ]]; then
     cd ..
     parallel --no-notice ${remote_args} --wd kickstart-tests --jobs ${TEST_JOBS:-4} \
              sudo PYTHONPATH=$PYTHONPATH scripts/launcher/run_one_test.py \
-                                                               -i ${_IMAGE} \
+                                                               -i ../install_images/${_IMAGE} \
```
jkonecny12: Is this working correctly. You are removing `${_IMAGE}` on line 321 but then using it here few lines later. Is this tested properly?

M4rtinK: The find on line 321 is an inverted find:

`find ../install_images -type f ! -name ${_IMAGE} -delete`

It will delete all files other than the one named `${_IMAGE}` and is used to cleanup any old images from the image storage folder.

As for testing, I've run multiple all-tests run with this and many test runs on the packaging tests group. I've been running all these with TEST_REMOTES_ONLY=yes (eq. no tests from the batch running on the machine that started the batch), so it might be good to tests the other usecase as well.

jkonecny12: Oh, OK then I guess it should be fine. :)

scripts/run_kickstart_tests.sh
M4rtinK: This is to support the `TEST_REMOTES_ONLY` != yes usecase, where also the machine that started the run is running some of the tests locally. This way you don't need to have a dedicated management machine/VM and all machines can run tests.

The side effect of that is that some things need to be done twice, once for all the remotes and once for the local machine, if `TEST_REMOTES_ONLY` != yes. The resulting environment then needs to be substantially similar so that run_one_tests.py can run the same on the remotes and one the local machine.

As for the removed rm call - we now copy the image file ti install_images folder instead of directly to the kickstart-tests folder:
```
-    cp ${IMAGE} ${_IMAGE}
+    mkdir -p ../install_images
+    cp ${IMAGE} ../install_images/
```
Therefore the rm call on `${_IMAGE}` is no longer relevant as `${_IMAGE}` will not be there).

PR #135, opened 2018-06-14
Fix the kstest report generation

M4rtinK:


scripts/launcher/lib/virtual_controller.py
```diff
@@ -303,7 +303,7 @@ def run(self):
             self._prepare_and_run()
         except InstallError as e:
             log.error("ERROR: Image creation failed: %s", e)
-            return False
+            return e
```
jkonecny12: I think you want to raise the exception above or resturn just a `str(e)` instead.

scripts/launcher/run_one_test.py
```diff
@@ -115,8 +116,10 @@ def run_test(self):
 
         virt_manager = VirtualManager(v_conf)
 
-        if not virt_manager.run():
-            self._result_formatter.report_result(False, "Virtual machine installation failed.")
+        try:
+            virt_manager.run()
+        except InstallError as e:
+            self._result_formatter.report_result(False, e)
```
jkonecny12: You should call this with `str(e)` instead.

M4rtinK:
Updated the PR:
- changed exception propagation as suggested by jkonecny
- handle the two types of tracebacks that show up in the logs in the run report script

PR #136, opened 2018-06-18
Add ansible playbooks for remote tests

rvykydal:


ansible/roles/kstest/tasks/main.yml
```diff
@@ -0,0 +1,105 @@
...
+      - git
+      - libselinux-python
+      - NetworkManager
+      - lorax-lmc-virt # kstests livemedia-creator
+      - openssh-clients # kstests scp
+      - parallel # kstests
+      - python3-libvirt # kstests
+      - libguestfs-tools # kstests
+      - vim
+    state: latest
```
jkonecny12: Just from curiosity, doesn't this run multiple DNF transactions for every package one? That would result in a much slower run.
I guess it is not a problem I just wonder what will happen.

ansible/roles/kstest/tasks/main.yml
```diff
@@ -0,0 +1,105 @@
...
+    name: whell
+    state: present
+
+- name: Allow 'wheel' group to have passwordless sudo
+  lineinfile:
+    dest: /etc/sudoers
+    state: present
+    regexp: '^%wheel'
+    line: '%wheel ALL=(ALL) NOPASSWD: ALL'
+    validate: 'visudo -cf %s'
```
jkonecny12: This is enough for now but for future I would like to tighten this permissions.

ansible/roles/kstest-master/files/scripts/kstests_history.py
```diff
@@ -0,0 +1,120 @@
...
+</thead>
+<tbody>
+{}
+</tbody>
+</table>
+</body>
+</html>
+""".format(thead, tbody)
+
+print(page)
```
jkonecny12: I would prefer to have some structure here, functions at least. Because you know, no script is small enough to be written as a small... The point is this script has a potential to grow and it may grow in this manner so it will result in a really big script without any structure.
However, I know this is just a helper script so I won't block merge because of this.

ansible/roles/kstest/tasks/main.yml
rvykydal: Good point, looking here: https://github.com/ansible/ansible-modules-extras/issues/642
I think it will run just one transaction.

ansible/roles/kstest-master/files/scripts/kstests_history.py
rvykydal: I agree. The script needs some care but I don't want to spend time on it now. We may throw it completely away as well in near future. Perhaps it doesn't even belong to the playbook, which should just sync the results for further processing.

rvykydal:
I've added two more fixups.
As for the .gitignore - it is a workaround for git not allowing to add empty directory into repository.

ansible/roles/kstest/tasks/main.yml
jkonecny12: Good to know :)

ansible/roles/kstest-master/templates/run_tests.sh.j2
```diff
@@ -3,7 +3,7 @@
 
 set -x
 
-TEST_REMOTES="{{ kstest_test_remotes }}"
+TEST_REMOTES="{{ kstest_test_remotes | default(groups['kstest']|join(' ')) }}"
 COMPOSE_URL="{{ kstest_compose_url }}"
 TEST_JOBS={{ kstest_test_jobs }}
 KSTESTS="{{ kstest_tests_to_run }}"
```
jkonecny12: Remove period from this commit message header please.

ansible/roles/kstest-master/tasks/main.yml
```diff
@@ -52,6 +52,14 @@
...
     - schedule-test
 
+- name: Enable and start crond service
+  service:
+    name: crond
+    state: started
+    enabled: yes
+  tags:
+    - schedule-test
+
```
jkonecny12: The same here.

rvykydal:
I've fixed the messages added two small fixups:
- https://github.com/rvykydal/kickstart-tests/blob/add-ansible-playbooks-for-remote-tests/ansible/roles/kstest-master/defaults/main.yml#L6
- the last line in https://github.com/rvykydal/kickstart-tests/blob/add-ansible-playbooks-for-remote-tests/ansible/roles/kstest-master/README.md

PR #137, opened 2018-06-19
Implement first version of test_manager

jkonecny12:
The `test_manager` tool is collecting kickstart tests and test preparation.

It can be used as tool or a library.

This library will be used by `run_kickstart_tests.sh` after rewrite to python 3.


scripts/test_manager/collector.py
```diff
@@ -0,0 +1,108 @@
...
+            if group in t.metadata.groups:
+                if cls._test_failure_print_error(t):
+                    continue
+                else:
+                    result.add(t)
+
+        return result
+
+    @classmethod
+    def _test_failure_print_error(cls, test):
```
poncovka: This method name is a little confusing :-) Maybe `skip_known_failure`?

scripts/test_manager/collector.py
```diff
@@ -0,0 +1,108 @@
...
+        if test.metadata.known_failure:
+            print("Test {} is marked as known failure, skipping...".format(test.name))
+            return True
+
+        return False
+
+    @classmethod
+    def process_tests(cls, tests):
```
poncovka: Process tests can be anything. Maybe `find_by_paths`?

jkonecny12:
**UPDATED:**
* Add error handling on the test level. 
    * Multiple errors can be stored for each test.
    * So tests are storing errors raised in the process.
* Remove prints. 
    * Prints are now located only in `test_manager.py`.
    * Prints shouldn't be part of the library.

poncovka:
The error handling is a little confusing. It is hard to tell when you should add an error and when you should raise an error.

jkonecny12:
You should always add an error. Raising an error is only easier to stop processing of the test and then it is saved to test.

jkonecny12:
**UPDATED:**
* Rework test error handling.
* Errors are still saved to tests but consistently on one place.
* Running test configuration changed to filter like solution. Set filters before and just run these filters on a group of tests.

scripts/test_manager.py
```diff
@@ -0,0 +1,110 @@
...
+if __name__ == "__main__":
+    parser = ArgumentParser()
+    parser.parse()
+
+    run_known_failure_tests = False
+    collector = TestCollector()
+    tests = []
+    if parser.tests:
+        tests = collector.find_by_paths(parser.tests)
+        run_known_failure_tests = True
```
poncovka: This variable is not used anywhere.

PR #155, opened 2018-07-27
Make it possible to skip kickstart tests by type

M4rtinK:
Each test can have multiple types set by the TESTTYPE variable
and the skip option (-s) will skip such tests as long as at least
one test type matches the skip string specified.

### DIFF
--- a/scripts/run_kickstart_tests.sh
+++ b/scripts/run_kickstart_tests.sh
@@ -63,8 +63,9 @@ KEEPIT=${KEEPIT:-0}
 UPDATES_IMG=""
 
 TESTTYPE=""
+SKIP_TESTTYPE=""
 
-while getopts ":i:k:t:u:b:" opt; do
+while getopts ":i:k:t:s:u:b:" opt; do
     case $opt in
        i)
            # If this wasn't set from the environment, set it from the command line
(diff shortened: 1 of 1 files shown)


scripts/run_kickstart_tests.sh
```diff
@@ -153,9 +158,12 @@ function find_tests() {
     local tests=$(find . -maxdepth 1 -name '*sh' -a -perm -o+x)
 
     local newtests=""
+    local skipped_tests=""
     for f in ${tests}; do
         if [[ "$TESTTYPE" != "" && "$(grep TESTTYPE= ${f})" =~ "${TESTTYPE}" ]]; then
             newtests+="${f} "
+        elif [[ "$SKIP_TESTTYPE" != "" && "$(grep TESTTYPE= ${f})" =~ "${SKIP_TESTTYPE}" ]]; then
+            skipped_tests+="${f}"
```
poncovka: If you do this check first, it will be possible to also exclude tests from specified types of tests. For example, you could run all storage tests except for autopart with `-t storage -s autopart`.

jkonecny12: Good point, could you please make that change before merge.

M4rtinK:
Switched check order as suggested by @poncovka - thanks! :)

PR #165, opened 2018-08-06
Add some additional log based error checks

M4rtinK:


scripts/launcher/lib/log_handler.py
```diff
@@ -27,6 +27,18 @@ class VirtualLogRequestHandler(LogRequestHandler):
...
+    # of the default ones contained in Lorax
+    added_simple_tests = [
+        "Payload setup error:",
+        "Out of memory:",
+    ]
+
+    def __init__(self):
+        super().__init__()
+        # add any additional error lines
+        super().simple_tests.extend(self.added_simple_tests)
```
poncovka: Please, use `self` instead of `super()`.

PR #174, opened 2018-09-07
Add utility playbooks

M4rtinK:
Add playbooks for easy test runner management.

cleanup - clean stale tests results
download-image - download a new boot iso
list-images - list available test images
reboot - perform a reboot of the runners
update - update the system to latest package version via DNF

This playbook basically target the usecase where kickstart test
are used to validate development and ad-hoc component changes
where the test runners are not created on demand, but created
once and then used for multiple test runs as needed.

### DIFF
--- a/ansible/utility/cleanup.yml
+++ b/ansible/utility/cleanup.yml
@@ -1,0 +1,19 @@
+---
+- hosts:
+    kstest
+  become: true
+  become_user: root
+
+  tasks:
+  - name: Find kickstart test results
+    find:
+      paths: /var/tmp
+      patterns: "kstest-*"
+      file_type: any
+    register: wildcard_items_to_delete
+
+  - name: Femove kickstart test results
+    file:
+      path: "{{ item.path }}"
+      state: absent
+    with_items: "{{ wildcard_items_to_delete.files }}"
--- a/ansible/utility/download-image.yml
+++ b/ansible/utility/download-image.yml
@@ -1,0 +1,21 @@
+---
+# run with --extra-vars "image_url=<boot iso URL>"
+- hosts:
+    kstest
+  become: true
+  become_user: root
+
+  tasks:
+
+  - name: Create directory for install images
+    file:
+      path: /home/kstest/install_images/
+      state: directory
+      owner: kstest
+      group: kstest
+
+  - name: Download new boot iso
+    get_url:
+      url: "{{ image_url }}"
+      dest: /home/kstest/install_images/
+
--- a/ansible/utility/update-system.yml
+++ b/ansible/utility/update-system.yml
@@ -1,0 +1,11 @@
+---
+- hosts:
+    kstest
+  become: true
+  become_user: root
+
+  tasks:
+  - name: Update all packages
+    dnf:
+      name: "*"
+      state: latest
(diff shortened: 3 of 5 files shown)


ansible/utility/cleanup.yml
```diff
@@ -0,0 +1,25 @@
...
+    file:
+      path: "{{ item.path }}"
+      state: absent
+    with_items: "{{ wildcard_items_to_delete.files }}"
+
+
+
+    #shell: rm -rf kstest*
+    #args:
+    #  chdir: /var/tmp
```
jkonecny12: Please remove commented code.

ansible/utility/download-image.yml
```diff
@@ -0,0 +1,21 @@
...
+  - name: Create directory for install images
+    file:
+      path: /home/kstest/install_images/
+      state: directory
+      owner: kstest
+      group: kstest
+
+  - name: Download new boot iso
+    get_url:
+      url: "{{ url }}"
```
jkonecny12: `url` is too generic. Please rename that variable to avoid collision.

ansible/utility/update-system.yml
```diff
@@ -0,0 +1,11 @@
...
+- hosts:
+    kstest
+  become: true
+  become_user: root
+
+  tasks:
+  - name: Update all packages
+    dnf:
+      name: "*"
+      state: latest
```
jkonecny12: I don't think you need this playbook really. There is not much different call between:

`ansible-playbook -i hosts -k -K ./ansible/utility/update-system.yml`

and

`ansible kstest -i hosts -b -k -K  -m dnf -a "name=libguestfs-tools-c state=latest"`

ansible/utility/cleanup.yml
```diff
@@ -0,0 +1,25 @@
+---
+- hosts:
+    kstest
+  become: true
+  become_user: root
+
+  tasks:
+  - name: Ansible find kickstart test results
```
rvykydal: I'd remove "Ansible"

ansible/utility/cleanup.yml
```diff
@@ -0,0 +1,25 @@
...
+    find:
+      paths: /var/tmp
+      patterns: "kstest-*"
+      file_type: any
+    register: wildcard_items_to_delete
+
+  - name: Ansible remove kickstart test results
```
rvykydal: Same here.

ansible/utility/update-system.yml
rvykydal: Hm, I think in this case reducing amount of stuff to type (and perhaps recall) can make a difference. I don't have problem having the playbook there if anyone finds it useful. It is just a utility.

jkonecny12: I agree with the last part. I don't see it useful but I don't have a problem to leave it there if anyone finds it useful.

ansible/utility/cleanup.yml
M4rtinK: Cleaned up. :)

ansible/utility/download-image.yml
M4rtinK: Changed to `image_url`.

ansible/utility/cleanup.yml
M4rtinK: OK.

ansible/utility/cleanup.yml
M4rtinK: Also removed.

PR #176, opened 2018-09-13
Add host id for kickstart test result line

M4rtinK:
When many kickstart tests are running in parallel on different
machines/VM hosts it can be very useful to know on which host
a given test was run.

This information can be used for example to debug host specific issues
that break tests (machine is too slow or unstable, has broken
networking, etc.) or to directly access test logs on machines during a
long run.

So add a host id (at the moment simply the machine hostname) to
the RESULT line, so we can easily see on which machine did the test run.

Also adjust the run_report.sh script to be compatible with the new RESULT format.

### DIFF
--- a/scripts/launcher/run_one_test.py
+++ b/scripts/launcher/run_one_test.py
@@ -35,6 +35,7 @@
 import os
 import shutil
 import subprocess
+import socket
 
 from lib.temp_manager import TempManager
 from lib.configuration import RunnerConfiguration, VirtualConfiguration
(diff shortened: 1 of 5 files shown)


scripts/launcher/run_one_test.py
```diff
@@ -55,11 +56,22 @@ def __init__(self, configuration, tmp_dir):
...
         self._shell = ShellLauncher(configuration, tmp_dir)
-        self._result_formatter = ResultFormatter(self._conf.ks_test_name)
+        self._result_formatter = ResultFormatter(self._conf.ks_test_name, host_id=self.host_id)
         # test prepare function can change place of the kickstart test
         # so the validator will be set later
         self._validator = None
 
+    @property
+    def host_id(self):
+        """Return a show string identyfying the host where the test is running.
```
jkonecny12: Please fix grammar 'identyfying' -> 'identifying'.

PR #178, opened 2018-09-19
ansible: adapt summary script to adding host into result report

rvykydal:
Should parse both old and new result reports.


rvykydal:
Related to this change: https://github.com/rhinstaller/kickstart-tests/pull/176

PR #186, opened 2018-10-08
Update libusbx for new qemu (#1625641)

rvykydal:
Until this is resolved in qemu.


jkonecny12:
I wouldn't recommend to update all packages. We could have more problems in the future.

PR #187, opened 2018-10-09
Fix network-noipv4-httpks test for rhel-devel

rvykydal:


rvykydal:
The PR is obsoleted by https://github.com/rhinstaller/kickstart-tests/pull/336

PR #189, opened 2018-10-10
Fixes for various kickstart tests

M4rtinK:
- fix the Fedora live image test (bigger storage & write SUCCESS to result file)
- make the container test authselect aware
- catch some additional & new log messages in monitor

### DIFF
--- a/scripts/launcher/lib/log_handler.py
+++ b/scripts/launcher/lib/log_handler.py
@@ -32,10 +32,12 @@ class VirtualLogRequestHandler(LogRequestHandler):
     simple_tests = LogRequestHandler.simple_tests + [
         "Payload setup error:",
         "Out of memory:",
-        "The following group or module is missing:",
+        "Would you like to ignore this and continue with installation?",
+        "Some packages, groups or modules are broken, the installation will be aborted.",
         "Stream was not specified for a module",
         "The following problem occurred on line",  # kickstart parsing error
         "storage configuration failed:",
+        "Not enough space in file systems for the current software selection.",
     ]
 
     def iserror(self, line):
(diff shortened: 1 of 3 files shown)


scripts/launcher/lib/log_handler.py
```diff
@@ -32,10 +32,12 @@ class VirtualLogRequestHandler(LogRequestHandler):
     simple_tests = LogRequestHandler.simple_tests + [
         "Payload setup error:",
         "Out of memory:",
-        "The following group or module is missing:",
+        "Would you like to ignore this and continue with installation?",
+        "Some packages, groups or modules are broken, the installation will be aborted.",
         "Stream was not specified for a module",
         "The following problem occurred on line",  # kickstart parsing error
         "storage configuration failed:",
+        "Disk Requirements:",  # payload does not fit into available storage
```
jkonecny12: This is really generic message. Couldn't this make trouble to us in the future?

M4rtinK: IIRC there was no better single line message to match on unfortunately. 
So I suggest to merge this now and then go change the message in Anaconda to something more distinct (like "Can't Satisfy Disk Requirements:"). Then once it bubbles it's way through on Fedora & RHEL we can update it here as well. :)

M4rtinK: Actually it looks like the message comes from DNF:
https://github.com/rpm-software-management/dnf/blob/e9f362facbaf999348e56b5f6a7bb4c245a1080a/dnf/base.py#L880

But this seems to be prefixed with "Error Summary", which could be a better match.

M4rtinK:
Found a more specify message for out of storage space errors. :) That clears the remaining issue, so merging now. :)

PR #194, opened 2018-10-24
Make it possible to skip multiple test types

M4rtinK:
Till now it was possible to specify a single test type to run
and also skip a single test type.

This might not be enough if one for example wants to run
all the packaging related tests but also skip all tests marked
as both knowfailure and rhel-only.

So make it possible to pass multiple test types to skip via the
run_kickstart_tests.sh script -s option.

Please note that due to specifics of the shell option parser
multiple test types need to be passed as a quoted space delimited
string:

-s "rhel-only knownfailure"

### DIFF
--- a/scripts/run_kickstart_tests.sh
+++ b/scripts/run_kickstart_tests.sh
@@ -63,7 +63,7 @@ KEEPIT=${KEEPIT:-0}
 UPDATES_IMG=""
 
 TESTTYPE=""
-SKIP_TESTTYPE=""
+SKIP_TESTTYPES=""
 
 while getopts ":i:k:t:s:u:b:" opt; do
     case $opt in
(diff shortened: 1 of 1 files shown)


scripts/run_kickstart_tests.sh
```diff
@@ -85,8 +85,11 @@ while getopts ":i:k:t:s:u:b:" opt; do
            TESTTYPE=$OPTARG
            ;;
        s)
-           # Exclude tests of the given testtype. We'll do a pretty stupid test for it.
-           SKIP_TESTTYPE=$OPTARG
+           # Exclude tests of the given testtypes. Multiple test types can be specified
```
rvykydal: I'd use "testtypes" -> "test types" or "TESTTYPEs". But I wouldn't insist.

scripts/run_kickstart_tests.sh
```diff
@@ -163,7 +178,7 @@ function find_tests() {
     local newtests=""
     local skipped_tests=""
     for f in ${tests}; do
-        if [[ "$SKIP_TESTTYPE" != "" && "$(grep TESTTYPE= ${f})" =~ "${SKIP_TESTTYPE}" ]]; then
+        if should_skip_test ${f}; then
             skipped_tests+="${f}"
```
rvykydal: I was just wondering why the variable when it is not used / passed anywhere.

M4rtinK: Previously I planned to print the skipped tests as well as the selected tests but this turns out to be non trivial due to how bash functions work. It's pretty hard to return two lists from a single function and also functions apparently can't write to global variables. 

So let's just leave this as-is, we plan to replace this shell script with Python anyway.

PR #195, opened 2018-11-06
Make test easy to test and extend

jkonecny12:
These are just minor tweaks which will prepare ground for future work.

The most significant one is done in the configuration. Parser is separated from configuration object.


jkonecny12:
**UPDATED:**
* Renamed `confs` module to `conf`.
* Add some minor code clean-ups

PR #198, opened 2018-11-08
Make sure sane locale is used on remote hosts

M4rtinK:
We use GNU Parallel to run the kickstart test jobs both locally
and on remote hosts. When it runs the jobs Parallel apparently
imports shell variables from the local host starting the job batch
to the shell environment on the remote hosts running the jobs.

This can cause issues if there is a locale availability mismatch
between the local and remote hosts, such as local host having
locale cs_CZ.UTF-8 while the remote hosts are lightweight cloud images
that only have a basic set of locales available (C.UTF-8, en_US.UTF-8,
etc.).

This mismatch then apparently causes the shell environment to
fallback to a **non-UTF-8 C locale**, wreaking all sorts of havoc,
especially with with log monitoring in Lorax/LMC, as the logs
*do* contain Unicode and make the log monitor crash due to Unicode
encoding errors.

So to avoid this nasty edge case, always set a well-known locale
to the LANG variable before starting remote jobs with Parallel.

### DIFF
--- a/scripts/run_kickstart_tests.sh
+++ b/scripts/run_kickstart_tests.sh
@@ -370,6 +370,14 @@ if [[ "$TEST_REMOTES" != "" ]]; then
     cd ..
 
     echo "starting tests"
+
+    # Parallel aparently tends to import environment shell variables to the
+    # remote shell run environment, which can cause issues if the host that
+    # has initiated a test run has a locale the host running the tests does
+    # not have. So always set a well known locale, which should prevent this
+    # missmatch from happening.
+    export LANG=en_US.UTF-8
+
     parallel --no-notice ${remote_args} --wd kickstart-tests --jobs ${TEST_JOBS:-4} \
              sudo PYTHONPATH=$PYTHONPATH scripts/launcher/run_one_test.py \
                                                                -i ../install_images/${_IMAGE} \


scripts/run_kickstart_tests.sh
```diff
@@ -370,6 +370,14 @@ if [[ "$TEST_REMOTES" != "" ]]; then
...
 
     echo "starting tests"
+
+    # Parallel aparently tends to import environment shell variables to the
+    # remote shell run environment, which can cause issues if the host that
+    # has initiated a test run has a locale the host running the tests does
+    # not have. So always set a well known locale, which should prevent this
+    # missmatch from happening.
+    export LANG=en_US.UTF-8
+
```
jkonecny12: This will work but the issue is not fixed. What is the test machines would have other `LANG` than `en_US` one?

M4rtinK: I'm afraid that in such a case all bets are off - it means you have to explicitly request UTF-8 encoding in every place you ingest or output any string data. This includes all string data reading/writing from sockets and files and also log handlers. I was able to get it work, but it required adding the explicit encoding requests in many places both in Lorax and kickstart test scripts and it was still not clear if it was enough. Also note that basically any software doing string processing on string that might contain unicode can fail (possibly even silently), even if you call it as from the command line.

So I think just making sure the workers where we run tests have en_US.UTF-8 available (even the Fedora cloud images have it out of the box) seems like the best bet at the moment rather than trying to hunt down all places where explicit encoding would have to be fixed for the broken no-UTF-8 locale case.

PR #199, opened 2018-11-09
Deduplicate boot option overrides & prevent Dracut emergency shell

M4rtinK:
If dracut fails to reach successful switchroot it will by default start
its emergency shell and wait for user input.

This is a problem for kickstart tests as all our error monitoring mechanisms
rely on log forwarding from the test VM, which only starts in stage2 after
a successful switchroot.

So add rd.shell=0 rd.emergency=poweroff to default command boot options to
prevent Dracut from entering the emergency shell and to shutdown
instead.

### DIFF
--- a/bindtomac-ifname-httpks.sh
+++ b/bindtomac-ifname-httpks.sh
@@ -23,7 +23,7 @@ TESTTYPE="network"
 
 kernel_args() {
     . ${tmpdir}/ks_url
-    echo vnc debug=1 inst.debug ifname=ifname0:52:54:00:12:34:5D ifname=ifname1:52:54:00:12:34:5E ip=ifname0:dhcp inst.ks=${ks_url}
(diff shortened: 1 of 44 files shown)


M4rtinK:
Good point/ Looking at the code we already source functions.sh so I guess we could put the default boot options into a variable and then use it as the base to which the tests that override boot options add their stuff.

`DEFAULT_BOOTOPTS="inst.text debug=1 inst.debug inst.anaconda_ping_timer=1 rd.shell=0 rd.emergency=poweroff"`

the in the override:

```
  kernel_args() {
     echo $DEFAULT_BOOTOPTS inst.nosave=input_ks
  }

```

M4rtinK:
I've updated the pull request to first also de-duplicate the boot option ovverrides as sugested by @rvykydal. :)

M4rtinK:
@jkonecny12 Well, this is just yet another incremental step towards a glorious future where there will be proper metadata per test case & robust non-bash runner that will make sure proper boot options are used for the given test run & environment. :)

PR #201, opened 2018-11-12
Add test for prefixdevname feature

rvykydal:
Note: this should be rhel-only test.

The test will fail until the bugs for prefixdevname are fixed.


rvykydal:
@jkonecny12 
The bugs are in prefixdevname, so I think we should enable the test now. In general I don't think we should disable a test because other component makes it failing.
Also, we could see in rhel nighlty tests results when the bugs are resolved instead of polling the status of BZs and enabling the test then.

PR #208, opened 2018-11-16
linchpin: add status --show-inventory option

rvykydal:
I find it handy when working with long-term running target.  Also moved showing
the inventory path from simple status command which is targeted at the status
of running test to status --show-inventory.

### DIFF
--- a/kstests-in-cloud.sh
+++ b/kstests-in-cloud.sh
@@ -83,7 +83,7 @@ Options:
 
     --remote-user            remote user for deployment of provisioned runners by ansible;
                              for example for Fedora cloud images it is "fedora";
-                             for RHEL could images it is "cloud-user"
+                             for RHEL cloud images it is "cloud-user"
 
   Test configuration options ("run" command):
 
--- a/kstests-in-cloud.sh
+++ b/kstests-in-cloud.sh
@@ -28,6 +28,8 @@ WHEN=""
 LOGFILE=""
 VIRTUALENV_PATH=""
 
+STATUS_MODE=""
+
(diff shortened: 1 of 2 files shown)


kstests-in-cloud.sh
```diff
@@ -79,7 +81,7 @@ Options:
 
     --remote-user            remote user for deployment of provisioned runners by ansible;
                              for example for Fedora cloud images it is "fedora";
-                             for RHEL could images it is "cloud-user"
+                             for RHEL cloud images it is "cloud-user"
```
jkonecny12: This shouldn't be part of this commit. I know it is a nitpick but it creates not-useful noise in a history and similar.

kstests-in-cloud.sh
```diff
@@ -112,7 +118,7 @@ Options:
 HELP_USAGE
 }
 
-options=$(getopt -o k:r:c:p: --long cloud:,results:,key-name:,key-use-existing,key-upload:,ansible-private-key:,key-use-for-master,test-configuration:,pinfile:,when:,remove,logfile:,scheduled,remote-user:,virtualenv:,ansible-python-interpreter: -- "$@")
+options=$(getopt -o k:r:c:p: --long cloud:,results:,key-name:,key-use-existing,key-upload:,ansible-private-key:,key-use-for-master,test-configuration:,pinfile:,when:,remove,logfile:,scheduled,remote-user:,virtualenv:,ansible-python-interpreter:,show-inventory -- "$@")
```
jkonecny12: Can't you please wrap this line? It is really hard to read and it gets bigger and bigger.

PR #212, opened 2019-01-10
Add initial RHEL8 platform configuration

M4rtinK:
Users are expected to use appropriate BaseOS and AppStream URLs
for RHEL8 in the repos/default.ks fragment.


fragments/platform/rhel8/repos/default.ks
```diff
@@ -0,0 +1,3 @@
+# Default Fedora Rawhide repositories
+url BASE_OS_REPO_URL_GOES_HERE
+repo --name=appstream APPSTREAM_REPO_URL_GOES_HERE
```
jkonecny12: We can do this or we can use `unified` solution. The second sounds easier to me.

M4rtinK: Sounds, good, I'll add one more comment line.

PR #217, opened 2019-01-23
Small linchpin improvements

jkonecny12:
Attempt to make usage of the linchpin part of the kickstart tests more obvious.

### DIFF
--- a/kstests-in-cloud.sh
+++ b/kstests-in-cloud.sh
@@ -73,11 +73,12 @@ Options:
     -k, --key-name NAME      name of the ssh key used for provisioning in cloud;
                              by default new key is generated on the cloud provider
     --key-use-existing       use the existing key --key-name from cloud
-    --key-upload PATH        upload public ssh key defined by PATH (as --key-name if defined)
+    --public-key-upload PATH upload public ssh key defined by PATH (as --key-name if defined)
     --ansible-private-key PATH
                              path to private ssh key to be used for ansible deployment;
                              if not defined generated key or user's default ssh key is used;
-                             this option may be required with --key-use-existing or --key-upload
+                             this option may be required with --key-use-existing or
+                             --public-key-upload
     --key-use-for-master     use the deployment key also as master runner key which
                              is used by master to access other test runners;
                              without the option a new temporary master runner key is generated;
@@ -128,7 +129,7 @@ Options:
 HELP_USAGE
 }
 
-options=$(getopt -o k:r:c:p: --long cloud:,results:,key-name:,key-use-existing,key-upload:,\
+options=$(getopt -o k:r:c:p: --long cloud:,results:,key-name:,key-use-existing,public-key-upload:,\
 ansible-private-key:,key-use-for-master,test-configuration:,pinfile:,when:,remove,logfile:,\
 scheduled,remote-user:,virtualenv:,ansible-python-interpreter:,test-run-timeout:,show-inventory,\
 force -- "$@")
@@ -149,14 +150,14 @@ while true; do
         ;;
     --key-use-existing)
         [[ ${KEY_MODE} == "upload" ]] && {
-            echo "Only one of --key-use-existing or --key-upload options can be used."
+            echo "Only one of --key-use-existing or --public-key-upload options can be used."
             exit 1
         }
         KEY_MODE=existing
         ;;
-    --key-upload)
+    --public-key-upload)
         [[ ${KEY_MODE} == "existing" ]] && {
-            echo "Only one of --key-use-existing or --key-upload options can be used."
+            echo "Only one of --key-use-existing or --public-key-upload options can be used."
             exit 1
         }
         shift;
@@ -269,7 +270,7 @@ if [[ ${COMMAND} == "schedule" ]]; then
             USE_KEY_FOR_MASTER_ARG="--key-use-for-master"
         fi
         if [[ ${KEY_MODE} == "upload" ]]; then
-            KEY_MODE_ARG="--key-upload"
+            KEY_MODE_ARG="--public-key-upload"
         elif [[ ${KEY_MODE} == "existing" ]]; then
             KEY_MODE_ARG="--key-use-existing"
         fi
--- a/kstests-in-cloud.sh
+++ b/kstests-in-cloud.sh
@@ -125,6 +125,23 @@ Options:
                              use alternative python interpreter on deployed hosts;
                              (for example /usr/bin/python3, /usr/libexec/platform-python)
 
+USAGE:
+
+Basic usage examples are described in this section.
+
+To run kickstart tests on TARGET created by linchpin with generated ssh keys:
+
+./$(basename $0) --results <path_to_local_folder_to_save_logs> --remote-user <user_to_connect_on_provisioned_machines> test [TARGET]
+
+
+To run kickstart tests on TARGET with a ssh key already uploaded:
+
+./$(basename $0) --key-name <target_key_name> --key-use-existing --ansible-private-key <path_to_the_private_key_part> test TARGET
+
+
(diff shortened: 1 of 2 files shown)


kstests-in-cloud.sh
```diff
@@ -39,7 +39,7 @@ usage () {
 $(basename $0) [options] COMMAND TARGET
 
 Run kickstart tests on runners temporarily provisioned by linchpin in cloud.
-Linchpin target name TARGET is defined in PinFile (linchpin/PinFile).
+Linchpin target topology name TARGET is defined in PinFile. See `topology_name` value in the linchpin/PinFile.
```
rvykydal: TARGET is actually target name. See the PinFile - target consists of topology (with the same name in default PinFile) and layout. Name of the target is defined in PinFile as the top-level element (not topology_name) - see example PinFiles, eg https://github.com/rhinstaller/kickstart-tests/blob/master/linchpin/examples/example1/PinFile which contains target example1. (There can be multiple targets defined in single PinFile).

But I agree I should improve this to make clear what target actually is - a group of resources (test runners) that can be provisioned / used / destroyed by the script, defined in PinFile. And mention that only single instance of target named TARGET can be running on single host at the same time.

kstests-in-cloud.sh
```diff
@@ -68,11 +68,12 @@ Options:
     -k, --key-name NAME      name of the ssh key used for provisioning in cloud;
                              by default new key is generated on the cloud provider
     --key-use-existing       use the existing key --key-name from cloud
-    --key-upload PATH        upload public ssh key defined by PATH (as --key-name if defined)
+    --public-key-upload PATH upload public ssh key defined by PATH (as --key-name if defined)
```
rvykydal: Please update also documentation: README.md and maybe examples README.md

kstests-in-cloud.sh
```diff
@@ -109,11 +110,28 @@ Options:
...
+
+
+To run kickstart tests on TARGET with a ssh key already uploaded:
+
+./$(basename $0) --key-name <target_key_name> --key-use-existing --ansible-private-key <path_to_the_private_key_part> test TARGET
+
+
+To run kickstart tests on TARGET with uploading your local ssh key:
+
+./$(basename $0) --public-key-upload <path_to_the_public_key> --ansible-private-key <path_to_the_private_key_part> test TARGET
```
rvykydal: Maybe it is good to also name the uploaded key?
`
To run kickstart tests on TARGET with uploading your local ssh key under name NAME:

./$(basename $0) --public-key-upload <path_to_the_public_key> --ansible-private-key <path_to_the_private_key_part> --key-name NAME test TARGET
`

kstests-in-cloud.sh
jkonecny12: I've tried to describe this better.

jkonecny12:
*UPDATED:*
* Implemented fixes proposed by @rvykydal . Thanks!

linchpin/README.md
```diff
@@ -125,3 +125,10 @@ Examples
 --------
 
 [examples](examples)
+
+Hints
+------
+
+Linchpin contains lot of ansible code and some of those have hidden secrets which makes failed output *censured*. That is great to avoid leaking passwords and others but it is really bad if you need to find out what just happens on the remote machine. To solve this situation you can enable [Ansible debug](https://docs.ansible.com/ansible/latest/reference_appendices/config.html?highlight=ansible_debug#envvar-ANSIBLE_DEBUG) mode.
```
rvykydal: I'd prefer something shorter, like:

"For more verbose *logging* from ansible modules used by linchpin you can enable [Ansible debug](https://docs.ansible.com/ansible/latest/reference_appendices/config.html?highlight=ansible_debug#envvar-ANSIBLE_DEBUG) mode."

kstests-in-cloud.sh
rvykydal: "See the value of the top-level element in the linchpin/PinFile for the name" ?

kstests-in-cloud.sh
```diff
@@ -71,6 +71,8 @@ Options:
                              the file is located in linchpin directory, default is "PinFile"
 
     -k, --key-name NAME      name of the ssh key used for provisioning in cloud;
+                             this applies to an existing (already upleaded) key or a new
+                             key you want to upload (see --public-key-upload)
```
rvykydal: It applies to a newly generated key as well - it will be named NAME

kstests-in-cloud.sh
```diff
@@ -71,6 +71,8 @@ Options:
                              the file is located in linchpin directory, default is "PinFile"
 
     -k, --key-name NAME      name of the ssh key used for provisioning in cloud;
+                             this applies to an existing (already upleaded) key or a new
```
rvykydal: "upleaded" -> "uploaded"

kstests-in-cloud.sh
```diff
@@ -125,6 +125,23 @@ Options:
...
+
+
+To run kickstart tests on TARGET with a ssh key already uploaded:
+
+./$(basename $0) --key-name <target_key_name> --key-use-existing --ansible-private-key <path_to_the_private_key_part> test TARGET
+
+
+To run kickstart tests on TARGET with uploading your local ssh key:
+
+./$(basename $0) --public-key-upload <path_to_the_public_key> --ansible-private-key <path_to_the_private_key_part> test TARGET
```
rvykydal: I'd like to be consistent with examples where I supply action and TARGET before options, ie
```
./$(basename $0) test TARGET --results <path_to_local_folder_to_save_logs> --remote-user <user_to_connect_on_provisioned_machines>
```
which seems more comprehensible to me. But it would not be consistent with the usage above:
```
$(basename $0) [options] COMMAND TARGET
```
So let's leave it as it is.

kstests-in-cloud.sh
jkonecny12: Sounds great. Thanks!

jkonecny12: Fixed.

linchpin/README.md
jkonecny12: Changed to smaller version with keeping the censured part.

kstests-in-cloud.sh
jkonecny12: Fixed.

kstests-in-cloud.sh
jkonecny12: Fixed by the above

kstests-in-cloud.sh
jkonecny12: OK

jkonecny12:
**UPDATED:**
* Fixed issues found by @rvykydal

kstests-in-cloud.sh
rvykydal: I dndn't mean to use this phrase as it is, I'd go with
"this applies to the name of a newly generated key (default), or to an already existing key that should be used ( --key-use-existing), or to the name under which a key will be uploaded (--public-key-upload)"

linchpin/README.md
rvykydal: Looks okay, just use *censored* instead of *censured*.

jkonecny12: True, fixed.

kstests-in-cloud.sh
jkonecny12: ohh, that you meant by the first comment

jkonecny12: Rephrased.

jkonecny12:
**UPDATED**
What do you think about the new version now?

kstests-in-cloud.sh
rvykydal: I'd change "or to" -> "or" but I'll create a separate PR for that so we can finally merge this PR

PR #224, opened 2019-03-05
Increase default provisioning ssh ping timeout

rvykydal:


rvykydal:
Closing, we don't seem to need it anymore for the cloud we are using now.

PR #227, opened 2019-03-05
update network gui connections tests for use of network module

rvykydal:


rvykydal:
This will fix networking kickstart tests failing after connecting of UI to network module (majority of network tests is failing).

PR #230, opened 2019-04-03
fix network gui connections test library function

rvykydal:
The syntax error resulted in:

/tmp/ks-script-zsjsncui: line 77: syntax error near unexpected token `fi'
/tmp/ks-script-zsjsncui: line 77: `    fi'

and passing of the test.


jkonecny12:
jenkins, test this please

PR #235, opened 2019-04-11
Fix running kstests from boot.iso with virt-install 2.1.0

M4rtinK:
Note, this *requires* virt-install 2.1.0, older versions of virt-install available in F28 and earlier will not work with it.

### DIFF
--- a/scripts/launcher/lib/virtual_controller.py
+++ b/scripts/launcher/lib/virtual_controller.py
@@ -157,7 +157,7 @@ class VirtualInstall(object):
             args.append(extra_args)
 
             args.append("--location")
-            args.append(self._iso.mount_dir)
+            args.append(self._iso.iso_path + ",kernel=isolinux/vmlinuz,initrd=isolinux/initrd.img")
 
         channel_args = "tcp,host={0}:{1},mode=connect,target_type=virtio" \
                        ",name=org.fedoraproject.anaconda.log.0".format(


scripts/launcher/lib/utils/iso_dev.py
```diff
@@ -47,6 +47,26 @@ def mount_dir(self):
                 raise IsoDeviceException("To read mount dir you have to mount the ISO first.")
         return self._iso_mount.mount_dir
 
+    @property
+    def kernel(self):
```
poncovka: These two properties are not used, so you can remove them.

jkonecny12: I don't see a purpose of the first commit, is that a leftover from other changes?

rvykydal:
@M4rtinK 
Martin, can you please update the PR for the remarks? We are ready to run the tests on F29 hypervisor.

scripts/launcher/lib/utils/iso_dev.py
M4rtinK: Turned out to be not needed in the end and while useful can still be added at any time, so we can definitely drop it.

M4rtinK:
> @M4rtinK
> Martin, can you please update the PR for the remarks? We are ready to run the tests on F29 hypervisor.

PR updated. :)

PR #237, opened 2019-04-12
ansible: Updates for Fedora 29

rvykydal:
Related to https://github.com/rhinstaller/kickstart-tests/pull/235
which makes us to move form F28 to F29 as test runner.

### DIFF
--- a/ansible/roles/kstest/tasks/main.yml
+++ b/ansible/roles/kstest/tasks/main.yml
@@ -24,6 +24,11 @@
       - rsync
     state: latest
 
+- name: Update all packages
+  dnf:
+    name: "*"
+    state: latest
+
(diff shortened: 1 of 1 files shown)


ansible/roles/kstest/tasks/main.yml
```diff
@@ -24,6 +24,11 @@
       - rsync
     state: latest
 
+- name: Update all packages
+  dnf:
+    name: "*"
+    state: latest
```
jkonecny12: When is this called?
Is it called only when a host is created or every time we are launching a test? I guess the first one is correct but want to be sure.

rvykydal: Yes, the first one is true.

rvykydal:
Commit message fixed

PR #238, opened 2019-04-15
Enable compression when transferring results via scp

M4rtinK:
Pass the -C flag to scp when transferring the results from
remote kickstart test workers, which should enabled compression
on the SSH channel used by scp.

This should not make anything slower but might speed up things a bit
as many of the artifacts are plain text log files which should
compress quite well.


M4rtinK:
I would kinda assume that part of the documentation might not be very up to date as it refers to modem lines, which unless we are talking satellite modems, are not in wide use anymore. In the age of modems the added compression overhead could have mattered quite a bit on the old CPUs of that time. But in the present age I doubt running gzip the result recovery stream would add any noticeable overhead & we transfer quite a bit of plain text and similar highly compressible artifacts when transferring the test results.

On the other hand, it could be a good idea to add some timing to compare two substantial test runs to see if there is a time difference. It could be there is none and adding the option is not needed.

poncovka:
I guess that this PR can be closed.

PR #241, opened 2019-04-17
Network device vars

rvykydal:
The PR is basically moving definition of network device names used by the tests into single place so it can be updated easily.
For example for running tests on Fedora 29 hypervisor where rhel and fedora is using different naming scheme, for RHEL 8 variant of tests it would mean changing the network-device-names.cfg file.

Related to https://github.com/rhinstaller/kickstart-tests/pull/235


network-device-names.cfg
```diff
@@ -0,0 +1,9 @@
+PREFIX="ens"
+SUFFIX=""
```
jkonecny12: We are sourcing this to a main configuration file.
Please add `NET_` prefix to these variables to avoid collisions.

rvykydal: Good point, fixed.

rvykydal:
Tested running nightly running kickstart tests on F29. Seems to be OK and in par with F28 without the patch. I'll let it run one more time to be sure.

PR #253, opened 2019-05-15
linchpin: allow to configure suffix of test run results directory

rvykydal:
Shows up to be more useful than prefix in some cases. Ie prefix is useful for
classes of tests, suffix for variants/test modifications.

Example: next nightly is configured to be run with some updates image for testing, or with kickstart-tests branch having some update to test. Then it would be handy to easily configure the suffix to the test run so it is easily distinguishable in the generated results history.


rvykydal:
suffix

PR #254, opened 2019-05-15
Revert "Change default boot options"

rvykydal:
This reverts commit 45cdfcafcd1fe906f33dfe665db5a7cc16c768c1.

The tests with noninteractive gui have still significantly worse results than
the "vnc" which should be investigated and fixed before changing the default.

Also https://github.com/rhinstaller/anaconda/pull/1961 which partially fixes
the issues should be merged first.

### DIFF
--- a/functions.sh
+++ b/functions.sh
@@ -31,7 +31,7 @@ inject_ks_to_initrd() {
     echo "true"
 }
 
-DEFAULT_BASIC_BOOTOPTS="inst.noninteractive debug=1 inst.debug"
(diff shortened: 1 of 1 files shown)


M4rtinK:
Maybe we could just drop the noninteractive option for now instead of reintroducing back the VNC option ? VNC is really not a good mode for finding out what went wrong for test failures/stuck tests and setting a mode overrides the custom mode set in a couple of the NTP tests.

rvykydal:
I agree with @M4rtinK, closing this and opening https://github.com/rhinstaller/kickstart-tests/pull/255

PR #255, opened 2019-05-15
Do not default to noninteractive mode yet

rvykydal:
The tests with noninteractive gui have still significantly worse results than
the "vnc" which should be investigated and fixed before changing the default.

Also https://github.com/rhinstaller/anaconda/pull/1961 which partially fixes
the issues should be merged first.

### DIFF
--- a/functions.sh
+++ b/functions.sh
@@ -31,7 +31,7 @@ inject_ks_to_initrd() {
     echo "true"
 }
 
-DEFAULT_BASIC_BOOTOPTS="inst.noninteractive debug=1 inst.debug"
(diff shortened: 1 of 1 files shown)


rvykydal:
Let's give updated https://github.com/rhinstaller/anaconda/pull/1961 a try before merging this.

rvykydal:
The results with updated https://github.com/rhinstaller/anaconda/pull/1961 seem really good. One notable difference is in results of payload / modules related tests. Perhaps worth looking into.

rvykydal:
Based on the latest results I am closing this PR, we can keep the inst.noninteractive as the default.

PR #257, opened 2019-05-20
Test post_install_tools_disabled is set correctly

M4rtinK:
Test that the post_install_tools_disabled option is set correctly
in the user interaction config file:

- post_install_tools_disabled should be set to 1 only if
  --disable/--disabled flag is passed to the firstboot
  kickstart command
- should not be set in any other case

This is implemented via a shared bash function running from
a validate() method override as the user interaction config file
is only written out after regular %post scripts have been run.

### DIFF
--- a/initial-setup-default.sh
+++ b/initial-setup-default.sh
@@ -20,3 +20,22 @@
 TESTTYPE="initial-setup"
 
 . ${KSTESTDIR}/functions.sh
+. ${KSTESTDIR}/validate-lib-initial-setup.sh
+
+validate() {
+    # check IS is disabled via validation library function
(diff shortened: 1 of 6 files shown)


M4rtinK:
NOTE: The `initial-setup-default` will fail and that's actually good, as it helped us find an actual bug (`post_install_tools_disabled = 0` when `firstboot` command is not used).

We might want to wait with merging this PR until the above mentioned issue is fixed.

M4rtinK:
Once the following Anaconda PR is merged, all these Initial Setup tests should pass: 
https://github.com/rhinstaller/anaconda/pull/1986

PR #273, opened 2019-08-12
newtork bridge: add fallback using ip if brctl is missing

rvykydal:


jkonecny12:
Shouldn't it be better to use `ip` all the time? It will always be available.

PR #279, opened 2019-08-16
Do not use letters in MAC to avoid fails on letter case inconsistencies

rvykydal:


jkonecny12:
Shouldn't we test the behavior of the letters on the MAC address instead of removing this part from the tests?

rvykydal:
No, we shouldn't care. It can depend on where the ifcfg comes from (dracut / NM) - which can change and irrelevant for the test result.

jkonecny12:
In that case please add this information to the commit message. Then you are free to merge it.

PR #287, opened 2019-09-20
linchpin: add nightly tests configuration to examples

rvykydal:


rvykydal:
I'd prefer to keep them, they are useful for us and they don't hurt anybody else - they are just real world examples.

jkonecny12:
We really should not expose parts of our internal infrastructure anywhere, it's better to be over-protective than having to solve problems when it's there. It's not useful for outside people, if you replace all the links with example.com it will be more obvious that community can't use it.

M4rtinK:
I think we should be able to use the public links for the Fedora repos at least. And maybe the CentOS 8 repo links for the RHEL example ?

rvykydal:
Okay, I think you are right, these are not examples but real configurations which don't belong here.

PR #288, opened 2019-09-20
linchpin: update documentation - clouds.yml example

rvykydal:


jkonecny12:
I was not able to found these variables anywhere in the code so I guess these are used by linchpin directly. Could you please confirm these to me? Also don't we want to add a Linchpin documentation link somewhere near to give user easy way to study what these values are?

rvykydal:
It is actually openstack configuration used by linchpin.

rvykydal:
The user using some cloud (or linchpin provider in general) - in this case openstack - should know what values he needs for configuring the access - required settings depend also on the cloud configuration (as the change confirms) so I don't see the point in any further documentation. The config can be used as is by python openstack client (just located in ~/.config/openstack/clouds.yml). Linchpin just has its own default location.

PR #296, opened 2019-10-02
linchpin: fix using multiple values in kstest_skip_test_types (-s)

rvykydal:
Fixes passing of options with values containing spaces to
run_kickstart_tests.sh script.

### DIFF
--- a/ansible/roles/kstest-master/templates/run_tests.sh.j2
+++ b/ansible/roles/kstest-master/templates/run_tests.sh.j2
@@ -147,23 +147,11 @@ if [[ -n "${UPDATES_IMAGE}" ]]; then
 fi
 
 
-### Use additional boot options
-
-ADDITIONAL_BOOT_OPTIONS_ARG=""
-if [[ -n "${ADDITIONAL_BOOT_OPTIONS}" ]]; then
-    ADDITIONAL_BOOT_OPTIONS_ARG="-b \"${ADDITIONAL_BOOT_OPTIONS}\""
-fi
-
 ### Run tests of specified type
 
 TEST_TYPE_ARG=""
 if [[ -n "${TEST_TYPE}" ]]; then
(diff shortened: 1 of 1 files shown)


ansible/roles/kstest-master/templates/run_tests.sh.j2
```diff
@@ -183,7 +184,7 @@ fi
 ### Run the test
 
 RESULT_LOG="${RESULT_PATH}/${LOG_FILENAME}"
-scripts/run_kickstart_tests.sh -i ${BOOT_ISO} -k 1 ${UPDATES_IMAGE_ARG} ${ADDITIONAL_BOOT_OPTIONS_ARG} ${TEST_TYPE_ARG} ${SKIP_TEST_TYPES_ARG} ${PLATFORM_ARG} ${OVERRIDE_ARG} ${KSTESTS} 2>&1 | tee ${RESULT_LOG}
+eval scripts/run_kickstart_tests.sh -i ${BOOT_ISO} -k 1 ${UPDATES_IMAGE_ARG} ${ADDITIONAL_BOOT_OPTIONS_ARG} ${TEST_TYPE_ARG} ${SKIP_TEST_TYPES_ARG} ${PLATFORM_ARG} ${OVERRIDE_ARG} ${KSTESTS} 2>&1 | tee ${RESULT_LOG}
```
jkonecny12: Is this really required? If the spaces are the problem, what about just add "" around the variables?

rvykydal: Yeah, I tried that, but it is not not working because empty value (eg ADDITIONAL_BOOT_OPTIONS_ARG="") would produce an empty argument ('') in this case.

jkonecny12: Please add a comment with explanation why eval and ideally with the link on detailed explanation.

PR #298, opened 2019-10-24
Create a group that covers the basic installation cases

poncovka:
Use the group `coverage` to quickly test your changes for the most common
uses cases of the installation.

### DIFF
--- a/autopart-encrypted-1.sh
+++ b/autopart-encrypted-1.sh
@@ -17,7 +17,7 @@
 #
 # Red Hat Author(s): Vendula Poncova <vponcova@redhat.com>
 
-TESTTYPE="autopart storage"
+TESTTYPE="autopart storage coverage"
 
 . ${KSTESTDIR}/functions.sh
 
--- a/bond2-httpks.sh
+++ b/bond2-httpks.sh
@@ -17,7 +17,7 @@
 #
 # Red Hat Author(s): Radek Vykydal <rvykydal@redhat.com>
 
-TESTTYPE=${TESTTYPE:-"network"}
+TESTTYPE="${TESTTYPE:-"network"} coverage"
 
 . ${KSTESTDIR}/functions.sh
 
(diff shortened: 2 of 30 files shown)


poncovka:
The selection is open to a discussion. I have added all tests that seem to be relevant.

poncovka:
@M4rtinK, @rvykydal any suggestions?

M4rtinK:
> @M4rtinK, @rvykydal any suggestions?

I would recommend to add some user & root tests, maybe:
`user-multiple` - that tests root & multi user creation, including duplicates
`rootpw-lock` - tests root password locking works correctly

Also it would be good to have at least one test run in TUI mode (the default is AFAIK still GUI), which should be the case for `default-systemd-target-tui`, which forces TUI mode.

And the last suggestion - it would be good to check if live image creation still works via `fedora-live-image-build`, but be aware this test is quite resource intensive & could run for some time, so I leave this to your discretion. :)

rvykydal:
I would
- replace network-static with more comprehensive network-static-2-httpks
- maybe replace bond with onboot-activate
- if we care about having some of virtual device types covered, I'd use bond2-httpks instead of bond
- and if we can have more I'd add team-pre which would also partly cover dimension of configuration in %pre

rvykydal:
Just a side note - I'd like to review and consolidate network tests some time (we have more than 60 now I think).
It should be also possible to create a single "test as much as possible" test covering (merging) more existing tests creating many configurations of many devices, including various types of virt at the same time, but it may get fragile running such a big and (in reality hardly used) configuration.

poncovka:
Updated. It is now a total of 30 tests.

poncovka:
> It should be also possible to create a single "test as much as possible" test covering (merging) more existing tests creating many configurations of many devices, including various types of virt at the same time, but it may get fragile running such a big and (in reality hardly used) configuration.

@rvykydal, I was thinking about that too, but I don't think that it is worth it.

PR #299, opened 2019-11-08
Updates image support

rvykydal:
The purpose if this patch-set is adding support for running tests with automatically generated updates image from git repository.

To achieve that I had to obtain anaconda version in the tested iso before running the test (to get default base reference/tag for updates image) and I decided to use lorax-packages.log from the tested iso. I added the file to the test results so we can:
- Start to get the version for the summary page from the file instead of anaconda.log of the tests which does not have to be always available - eg if all tests fail earlier, or if log files are deleted to save space. This is done by one of the patches.
- Add package changes info to the summary page (eg bottom of the column) by easy update of the script generating the summary (this is for another PR later)


rvykydal:
TODO: add example

PR #302, opened 2019-11-19
ansible: add package diff info to summary page

rvykydal:
Seems like the time for the script consolidation / refactoring is coming, maybe next time...


jkonecny12:
I'm not sure is it a good idea to drag in DNF as dependency? At least it would be great to have that as optional not a hard dependency.

rvykydal:
> I'm not sure is it a good idea to drag in DNF as dependency? At least it would be great to have that as optional not a hard dependency. The same works for hawkey.

https://github.com/rhinstaller/kickstart-tests/pull/302/commits/f58f6c8ec1c56377f19ec2799c6d605ebd69c2db

PR #334, opened 2020-04-03
Add harddrive tests

jkonecny12:
Add tests for installation from harddrive.

They all work that in pre-section they will download Server DVD iso and place that in some form to one of the harddrives.

These tests could be resource demanding because each test is downloading DVD iso.

RHEL variants was not tested.

### DIFF
--- a/fragments/platform/rhel8/section-data/server-dvd-link.ks
+++ b/fragments/platform/rhel8/section-data/server-dvd-link.ks
@@ -1,0 +1,7 @@
+# Create a variable in the pre/post section with link to a RHEL DVD.
+# The curl tool will download this ISO and it will be processed later in the section.
+_LINK=<ADD_LINK_TO_RHEL-8_ISO_HERE>
+ISO_LOCATION="$(curl -L $_LINK | grep -Po "RHEL-8.*?-x86_64-dvd1.iso" | head -n 1)"
+ISO_LOCATION="${_LINK}/${ISO_LOCATION}"
+
+echo $ISO_LOCATION
--- a/harddrive-install-tree.sh
+++ b/harddrive-install-tree.sh
@@ -1,0 +1,31 @@
+#
+# Copyright (C) 2020  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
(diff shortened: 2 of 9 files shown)


fragments/platform/rhel8/section-data/server-dvd-link.ks
```diff
@@ -0,0 +1,7 @@
+# Create a variable in the pre/post section with link to a Fedora rawhide server DVD.
```
rvykydal, line 1: Fedora rawhide -> RHEL, or maybe just with "link to a server DVD"

rvykydal:
I was in doubts about adding common_no_storage_and_payload.ks fragment because obviously big common fragments like this would need to be kept in sync with the small fragments if there are any changes but for the current state of the code I think it is okay and can be reworked or consolidated later when the right time comes (maybe only after we really see some updates of the fragments).
So it is ok.

rvykydal:
If you want you can share the .sh code in a way similar to https://github.com/rhinstaller/kickstart-tests/blob/master/bridge-2devs-pre.sh (eg by sourcing harddrive-iso.sh).

fragments/platform/rhel8/section-data/server-dvd-link.ks
jkonecny12: Fixed.

jkonecny12:
**UPDATED:**
* Fixed wrong comment.
* Use one shell script for all.
* Increase size of the repo source hard drive to be able to contain RHEL repositories.

harddrive-install-tree.sh
```diff
@@ -0,0 +1,31 @@
...
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+# Red Hat Author(s): Jiri Konecny <jkonecny@redhat.com>
+
+TESTTYPE="packaging"
```
rvykydal, line 20: There is one tricky point here.
When sourcing this file by .sh files by the other hdd tests their TESTTYPE would be overriden (for example if it is marked by knownfailure). To prevent that there should be:
`TESTTYPE=${TESTTYPE:-"packaging"}`

jkonecny12: ohh, another bashism. Thanks a lot for pointing this!

jkonecny12:
**UPDATED:**
* Fixed issue found by @rvykydal

jkonecny12:
**UPDATED:**
* Correct TESTTYPE also for the last test

PR #338, opened 2020-06-01
Update rhel versions of ifname-httpks tests for rebase in 8.3

rvykydal:
The behaviour is now the same on Fedora and RHEL.


rvykydal:
Sorry for posting before testing, this PR needs more care

rvykydal:
Now the PR should be finished, and the tests are passing.

PR #339, opened 2020-06-01
Update kstest runner playbook for Fedora 32

rvykydal:
* a package name change
* NetworkManager is default so the code enabling the service is not needed
  (moreover services check fails on one service currently)


rvykydal:
DO NOT MERGE

rvykydal:
The nightly kickstart tests seem to be running good on F32 host.

PR #345, opened 2020-06-11
Copy out kickstart

VladimirSlavik:
This appears to do what we want, but further discussion much welcome. I wonder, would it be useful to grab the whole root folder?

### DIFF
--- a/functions.sh
+++ b/functions.sh
@@ -99,10 +99,13 @@ validate_RESULT() {
     #
     # logs from Anaconda - whole /var/log/anaconda/ directory is copied out,
     #                      this can be used for saving specific test output
-    # anaconda.coverage
-    # RESULT file from the test
-    for item in /root/anaconda.coverage \
-                /var/log/anaconda/      \
+    # original-ks.cfg - the kickstart used for the test
+    # anaconda-ks.cfg - the kickstart saved after installation, useful for
+    #                   debugging
+    # RESULT - file from the test
+    for item in /root/original-ks.cfg \
+                /root/anaconda-ks.cfg \
+                /var/log/anaconda/    \
                 /root/RESULT
     do
         copy_file "${args}" "${item}" "${disksdir}"
--- a/functions.sh
+++ b/functions.sh
@@ -103,11 +103,7 @@ copy_interesting_files_from_system() {
                 /var/log/anaconda/    \
                 /root/RESULT
     do
-        if [[ -z "$3" ]]; then
-            copy_file "${args}" "${item}" "${disksdir}"
-        else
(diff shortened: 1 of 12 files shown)


functions.sh
```diff
@@ -100,8 +100,11 @@ validate_RESULT() {
     # logs from Anaconda - whole /var/log/anaconda/ directory is copied out,
     #                      this can be used for saving specific test output
     # anaconda.coverage
-    # RESULT file from the test
+    # anaconda-ks.cfg - the kickstart saved after installation, useful for
+    #                   debugging
+    # RESULT - file from the test
     for item in /root/anaconda.coverage \
```
poncovka, line 106: Could you remove `anaconda.coverage`, please? The file no longer exists.

functions.sh
```diff
@@ -99,10 +99,13 @@ validate_RESULT() {
...
-    # RESULT file from the test
-    for item in /root/anaconda.coverage \
-                /var/log/anaconda/      \
+    # original-ks.cfg - the kickstart used for the test
+    # anaconda-ks.cfg - the kickstart saved after installation, useful for
+    #                   debugging
+    # RESULT - file from the test
+    for item in /root/original-ks.cfg \
+                /root/anaconda-ks.cfg \
+                /var/log/anaconda/    \
```
poncovka, line 108: Btw. I am afraid that this is not the only list of these files in kickstart tests.

VladimirSlavik: Oh, definitely. I just keep pushing as I work...

VladimirSlavik:
I can split the last commit into another PR if that helps with reviews.

functions.sh
```diff
@@ -85,28 +85,42 @@ copy_file() {
...
                 /root/RESULT
     do
-        copy_file "${args}" "${item}" "${disksdir}"
+        if [[ -z "$3" ]]; then
+            copy_file "${args}" "${item}" "${disksdir}"
+        else
+            copy_file "${args}" "${item}" "${disksdir}" 2>/dev/null
```
poncovka, line 109: Maybe we could always redirect stderr to /dev/null to simplify the code a little.

VladimirSlavik: It would be definitely nicer, but I am not sure if it's a good idea. The core problem is: This is normally called by the tests, and then you want to see the errors. But it is also called by the "nosave" tests, and I don't understand why these are different.

functions.sh
VladimirSlavik: Done.

VladimirSlavik:
I started a kickstart test coverage batch for this...

VladimirSlavik:
Tests succeeded, merging.

PR #349, opened 2020-07-10
Fix the test for the lang command

poncovka:
Don't check the dependencies of the langpacks-ru package, because they can
change. It is enough to check that the langpacks-ru package is installed on
the new system.


poncovka:
I have updated the commit message.

PR #352, opened 2020-09-07
Enforce the LVM partitioning when necessary

poncovka:
The change of the default partitioning scheme to Btrfs broke some of the
kickstart tests, so enforce the LVM partitioning scheme to fix them.


rvykydal (approved):
Looks good to me, thank you.

PR #353, opened 2020-09-07
Fix the test for NTP pools

poncovka:
Fix the test for NTP pools and extend it with an NTS server.


rvykydal (approved):
Looks good to me, thank you.

PR #359, opened 2020-09-23
update network noipv4 tests for kickstart applied in stage 2

rvykydal:
With processing of kickstart in stage 2 the code path of ordinary kickstart is
almost the same as for the configuration in %pre section, therefore the -httpks
and -pre versions of the test are converging to do the same checks. In the
future, when we are sure processing kickstart in stage2 has no fatal drawbacks
I am probably going to remove (or reduce to a few cases) the number of -pre
variants of networking tests.

Add a device with both --noipv4 and --noipv6 - in this case the connection is
activated.

Fix the case with --noipv4 only - in this case the connection is not activated
and is replaced by default autoconnection (on systems where default
autoconnections are enabled).

### DIFF
--- a/network-noipv4-httpks.ks.in
+++ b/network-noipv4-httpks.ks.in
@@ -3,7 +3,8 @@
 install
 
 network --device=@KSTEST_NETDEV2@ --noipv4 --no-activate --onboot=yes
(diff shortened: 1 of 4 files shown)


jkonecny12 (approved):
LGTM as far as I can tell.

PR #361, opened 2020-09-30
Add a kickstart test for the RPM OSTree payload

poncovka:
Use Fedora IoT for testing of the RPM OSTree payload.


jkonecny12 (approved):
Looks great to me and great idea how to test ostree!

PR #364, opened 2020-10-09
Enable bridged-bond-pre test which is no more knownfailure.

rvykydal:
The support for bridged bond was added in
https://github.com/rhinstaller/anaconda/pull/2861


jkonecny12 (approved):
Looks good to me.

PR #365, opened 2020-10-09
Fix network-bootopts-bootif-httpks test

rvykydal:
Update it with version for NM default autcoconnections turned on.


jkonecny12 (approved):
LGTM!

PR #368, opened 2020-10-09
Don't install Fedora Server Edition

poncovka:
The module tests don't require to install the environment Fedora Server Edition.


jkonecny12 (commented):
Looks good to me!

M4rtinK (approved):
Looks good to me as well, thanks! :)

IIRC, the reason this was originally included had something to do with making sure the modular repositories being included on the target system, which AFAIK is no longer an issues. So dropping it is certainly fine. :)

PR #370, opened 2020-10-09
Update the module-non-default-profiles test

poncovka:
Simplify the test and update versions and profiles of the installed modules.
Don't install the environment Fedora Server Edition, because it is not required.


M4rtinK (approved):
Looks good to me, thanks! :)

jkonecny12 (approved):
LGTM!

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #367, opened 2020-10-09
Add dependencies required to run mkdud.py and make-addon-pkgs.py

rvykydal:
Currently used by proxy-* and driverdisk-disk tests.

PR #362, opened 2020-10-09
README: Download boot.iso from official Fedora mirror, add missing scp test dependency

martinpitt:
There is no reason to use or expose the RedHat VPN internal mirror. Use
the official one, which automatically redirects to your closest mirror.

PR #371, opened 2020-10-13
Add missing genisoimage dependency

martinpitt:
scripts/Makefile.prereqs and ibft.sh both use isoinfo, so install it.

Also add the other recently added dependencies to
containers/runner/Dockerfile.

### DIFF
--- a/containers/runner/Dockerfile
+++ b/containers/runner/Dockerfile
@@ -9,9 +9,14 @@ RUN dnf -y update && \
     git \
     virt-install \
     libguestfs-tools \
+    genisoimage \
     lorax-lmc-virt \
     parallel \
-    python3-libvirt && \
+    python3-libvirt \
+    createrepo \
+    python3-rpmfluff \
+    squid \
+    openssh-clients && \
     dnf -y clean all
 
 ENV APP_ROOT=/opt/kstest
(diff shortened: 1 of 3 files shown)


poncovka (approved):

containers/runner/Dockerfile
```diff
@@ -9,9 +9,14 @@ RUN dnf -y update && \
...
     libguestfs-tools \
+    genisoimage \
     lorax-lmc-virt \
     parallel \
-    python3-libvirt && \
+    python3-libvirt \
+    createrepo \
+    python3-rpmfluff \
+    squid \
+    openssh-clients && \
```
VladimirSlavik, line 19: This is more than just genisoimage...?

martinpitt: Right, that's the "Also add the other recently added dependencies.." part of the commit message. See PR #367 and PR #362

martinpitt: I can split up the commits if you prefer, of course.

VladimirSlavik: Ah, sorry, I missed that :-/ That works, of course.

VladimirSlavik (approved):

PR #372, opened 2020-10-13
Handle a message about a detected degraded array

poncovka:
Move the validation of the installation logs to the VirtualLogRequestHandler
class. Kill the installation if there is a critical message with the exception
of a core dump and a detected degraded array.

Related: rhbz#1886809


jkonecny12 (approved):
Looks good to me.

PR #373, opened 2020-10-13
staging ground for test run improvements

martinpitt:
This is a dumping ground for introducing a safe, efficient, and easy way to run kickstart-tests locally and on our or public infrastructure. I'll regularly break this out into smaller PRs which are ready to land.

 - [x] PR #371 
 - [x] PR #375
 - [x] PR #377
 - [x] PR #378
 - [x] PR #382
 - [x] PR #383
 - [x] PR #385
 - [x] PR #386
 - [x] PR #388
 - [x] PR #390
 - [x] PR #391
 - [x] PR #392
 - [x] Work around virt-install creation race (sent fix to https://github.com/virt-manager/virt-manager/pull/175): PR #396
 - [x] Investigate libvirt storage pool race: https://bugzilla.redhat.com/show_bug.cgi?id=1894359: Fixed in https://github.com/virt-manager/virt-manager/pull/179 ; can we pull that fix into the image somehow?
 - [ ] Figure out why current infra breaks on symlinking boot.iso (commit b94e277562f8)

PR #377, opened 2020-10-15
Rewrite probe_boot_iso.sh to not require privileges

martinpitt:
* Use isoinfo, unsquashfs, and libguestfs to unpack the boot.iso
   Matroshka, instead of relying on mount and the host kernel. That also
   works in an unprivileged container.

 * Robustify cleanup with a trap instead of error exit paths.

 * Enable `set -eu` for script robustness, and drop the explicit error
   handlers.

 * Source os-release instead of parsing it with shell/grep.

As this is quite expensive (takes ~ 40 seconds on my 10 year old laptop), I also added a fast path for Fedora (which takes virtually no time). I realize that this is disputable, and I appreciate some feedback about it. I'm happy to remove this if you prefer, or  we change it to divine what the current "Rawhide" version number is (I didn't find a  good and obvious way how to do that).

This is split out from #373 to make it easier to review, and it's a rather independent change.

I tested this with a Fedora rawhide, 32, and RHEL 8 boot.iso:

```
[kstest@0bd53c2e5471 kickstart-tests]$ scripts/probe_boot_iso.sh ../data/images/boot.iso
NAME=fedora
VERSION=Rawhide
[kstest@0bd53c2e5471 kickstart-tests]$ scripts/probe_boot_iso.sh ../data/images/boot-f32.iso
NAME=fedora
VERSION=32
[kstest@0bd53c2e5471 kickstart-tests]$ scripts/probe_boot_iso.sh ../data/images/boot-r8.iso 
NAME=rhel
VERSION=8.4
```

Without the fastpath, the first output has `VERSION=34` instead. In that case it's identical to the current master version.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -217,8 +217,9 @@ The following environment variables are currently supported:
   contains a name of the OS. Possible names can be "fedora", "rhel".
 
 - KSTEST_OS_VERSION - This variable is read from the input boot.iso and it
-  contains version of the OS. For example Fedora 26 have
-  KSTEST_OS_VERSION = 26 and RHEL 7.3 have KSTEST_OS_VERSION = 7.3 .
+  contains version of the OS. For example Fedora 26 has
+  KSTEST_OS_VERSION = 26, Fedora rawhide has "Rawhide", and RHEL 7.3 has
+  KSTEST_OS_VERSION = 7.3 .
 
 Chapter 3. Sharing common code in kickstart (.ks.in) files
 ==========================================================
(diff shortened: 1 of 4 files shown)


VladimirSlavik:
> to divine what the current "Rawhide" version number is (I didn't find a good and obvious way how to do that).

See https://github.com/rhinstaller/anaconda/pull/2855 for some pointers into that fine mess.

jkonecny12 (approved):
Wow, this is a great improvement!

PR #378, opened 2020-10-15
Drop mount/root requirement from VirtualInstall

martinpitt:
Use `blkid` to determine the label of the given ISO file instead of
pylorax. This eliminates the entire ISO mounting, `IsoDev` class, loop
device setup, and privilege requirement.

We also don't need the `.stage2` flag, as probe_boot_iso.sh *only*
accepts ISOs which have a stage2. (If that weren't so, we could make
probe_boot_iso.sh output that fact as a variable, and pass it to
VirtualInstall).


poncovka (approved):
It looks great! Thanks!

rvykydal (approved):
Looks good to me.

jkonecny12 (approved):
Great improvement! Thanks!

scripts/launcher/lib/virtual_controller.py
```diff
@@ -145,8 +146,8 @@ def _prepare_args(self):
             extra_args += " inst.cmdline"
         if self._kernel_args:
             extra_args += " " + self._kernel_args
-        if self._iso.stage2:
-            extra_args += " stage2=hd:CDLABEL={0}".format(udev_escape(self._iso.label))
+
+        extra_args += " stage2=hd:CDLABEL={0}".format(udev_escape(self._label))
```
VladimirSlavik, line 150: Ditto here `inst.stage2=...`

scripts/launcher/lib/virtual_controller.py
```diff
@@ -132,10 +134,9 @@ def _prepare_args(self):
...
-        if self._iso.stage2:
-            disk_opts = "path={0},device=cdrom,readonly=on,shareable=on".format(self._iso.iso_path)
-            args.append("--disk")
-            args.append(disk_opts)
+        disk_opts = "path={0},device=cdrom,readonly=on,shareable=on".format(self._iso)
+        args.append("--disk")
+        args.append(disk_opts)
 
         if self._ks_paths:
             extra_args = "ks=file:/{0}".format(os.path.basename(self._ks_paths[0]))
```
VladimirSlavik, line 142: I think this should be rather `inst.ks=...` since we're working towards making the prefix mandatory.

VladimirSlavik (approved):
Thank you, this is great!

It might be advisable to also update the kernel parameters while you're touching this, see comments.

scripts/launcher/lib/virtual_controller.py
martinpitt: Ack, thanks for spotting! I locally ran the test, and it works fine with the change.

scripts/launcher/lib/virtual_controller.py
martinpitt: Dito.

martinpitt:
**UPDATE**: Added new commit to move away from deprecated kernel command line options. @VladimirSlavik , PTAL?

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
LGTM!

martinpitt:
Last night's RHEL-8 tests were successful again (after PR #377 and follow-up fix #381), so landing this one now to feed the next spoonful :grin:

PR #379, opened 2020-10-19
Revert "Don't copy boot.iso into output directory"

martinpitt:
This broke the tests in the current CI infra:

    The iso /var/tmp/kstest-firewall.1jlt8l1g/boot.kstest.iso is missing.

This reverts commit b94e277562f8cc547684d154c685add2a1caa355.

### DIFF
--- a/scripts/launcher/run_one_test.py
+++ b/scripts/launcher/run_one_test.py
@@ -33,6 +33,7 @@
 # 99 - Test preparation failed
 
 import os
+import shutil
 import subprocess
 import socket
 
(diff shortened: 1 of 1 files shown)


rvykydal (approved):
Looks good to me, thank you.,

martinpitt:
Some context: I only tested that in the containers/runner/ environment, and a local VM, both of which worked fine. I suspect some differing SELinux context or whatnot. But with the official infra not being easy to reproduce, let's postpone this optimization. (And this reinforces why containerization is good -- easy to reproduce and debug problems)

PR #382, opened 2020-10-23
Move network tests from virsh default network to user network

martinpitt:
This isolates the VMs from one another, alleviating the need to allocate
non-overlapping MAC addresses. It also works in unprivileged containers
(creating a bridge does not).

This also provides a completely predictable and stable 10.0.2.2 host IP
from the guest's POV, so replace the find-ip script with a constant.

----

Last night's rhel-8 results look good, so here comes the next spoonful of changes split out from #373 :wink: 

I tested these in the containers/runner container, but not on the official infra. If it's alright with you (and otherwise reviewed ok), I'll let the nightly cron job be the final judge of this.


VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal:
Interesting, I didn't think and even was aware about user networking in kvm.
Looking here: https://www.linux-kvm.org/page/Networking, I suppose we are fine with these points of the Use case
- You don't need to access your guest from the network or from another guest.
- You are ready to take a huge performance hit.

bond-vlan-httpks.sh
```diff
@@ -31,7 +31,7 @@ kernel_args() {
 prepare_network() {
     local tmpdir=$1
     local network=$(basename ${tmpdir})
-    echo "network:default"
+    echo "user"
     echo "network:${network}"
```
rvykydal, line 35: I wonder how are we going to deal with these dynamically created networks,
(which we create to prevent static ip collisions in multiple VMs /test intances run on single hypervisor (runner)). Perhaps something for the next step ? They could be probably rather simplified using the user networking.

rvykydal (approved):
I think we can give it a shot for this nightly rhel-8 run.

bond-vlan-httpks.sh
martinpitt: Oh thanks, this bit wouldn't work. I ran quite a few networking tests in #373, but not all of them. I'll look into converting this to a user network as well. Indeed that wouldn't block this PR as long as we still run the tests on the current infra, but it will break running the tests in a container, as you can't define your own networks there.

martinpitt:
Yay, the rawhide tests are back to "works" as well, that library fix finally landed. So now we have two sets of good results to check against. So I'll land this, and check tomorrow's results -- if it breaks, I'll send a revert, but then at least I have a list of regressions to investigate.

bond-vlan-httpks.sh
martinpitt: Done in PR #383

PR #383, opened 2020-10-26
Move network tests from virsh custom networks to user network

martinpitt:
As a follow-up to commit 35c55f7bd3, convert the custom networks as
well. SLIRP user networks are already isolated from one another, so
inside the guest we can already use any random IP configuration.


martinpitt:
I tested all of these locally, and aside from `network-bootopts-static-legacy-httpks` (which is knownfailure, and not supported in Fedora), they all pass. I also [updated the test list](https://github.com/rhinstaller/kickstart-tests/pull/373#issuecomment-708208161) in the "bring it all together" PR #373 .

jkonecny12 (approved):
Wow a great simplification.

rvykydal (approved):
Looks good to me, thank you.

PR #384, opened 2020-10-26
Fix the kickstart test of the default filesystem type

poncovka:
Split the test into two tests: one for Fedora Workstation and one for RHEL.


jkonecny12 (approved):
LGTM!

PR #385, opened 2020-10-27
Adjust network-noipv4* tests for QEMU user network

martinpitt:
Commit 9a5b5daf98 broke these tests as the user network behaves slightly
differently than libvirt's default network. In particular, it is fully
IPv6 enabled.

Fix the `check_device_has_ipv4_address` check to be always "no", as
that's what the configuration actually requests (it's not clear why it
previously got an IPv4 address despite disabling IPv4). Also adjust the
NM connection name for the enp4s0 device -- it's now the same as the
device name, not "Wired connection 2".


martinpitt:
That fixes two of the four regressions in http://10.43.136.2/trees/kstests/user-net/results/index.html

jkonecny12 (approved):
Looks good to me.

rvykydal (approved):
Looks good to me.

PR #386, opened 2020-10-27
Fix runner container, make it unprivileged, auto-build to quay.io

martinpitt:
The current runner container does not work, it fails on some "libvirt failing to talk to dbus" SNAFU. With these updates, I can build and run kickstart tests as per containers/runner/README.md.

Please run thorugh README.md, and check that you can build and use the container as advertised.

This PR should be save to land quickly, as we don't use it anywhere right now (and as mentioned above, it's broken anyway). This will  enable tricks like [running KS tests on Travis](https://travis-ci.com/github/martinpitt/kickstart-tests/builds/193409766), and more importantly, make it really easy and safe for a developer to run them locally.

This also sets up a workflow to automatically build and push the container to quay.io.

### DIFF
--- a/containers/runner/README.md
+++ b/containers/runner/README.md
@@ -23,9 +23,8 @@ Run a test in a container
 
 Build the container:
 ```
-sudo podman build -t kstest-runner .
+podman build -t kstest-runner .
 ```
-Note: we can't use rootless containers because of mounting of images in the contaier.
 
 Define the directory for passing of data between the container and the system (via volume):
 ```
@@ -35,7 +34,7 @@ export VOLUME_DIR=${PWD}/data
 Create subdirs for test inputs (installation image) and outputs (logs):
 ```
 mkdir -p ${VOLUME_DIR}/images
-mkdir -p ${VOLUME_DIR}/logs
+mkdir -p -m 777 ${VOLUME_DIR}/logs
 ```
 
 Download the test subject (installer boot iso):
@@ -45,7 +44,7 @@ curl -L https://download.fedoraproject.org/pub/fedora/linux/development/rawhide/
 
 Run the test:
 ```
-sudo podman run --env KSTESTS_TEST=keyboard -v ${VOLUME_DIR}:/opt/kstest/data:z --name last-kstest --privileged=true --device=/dev/kvm kstest-runner
+podman run --env KSTESTS_TEST=keyboard -v ${VOLUME_DIR}:/opt/kstest/data:z --name last-kstest --device=/dev/kvm kstest-runner
 ```
 Instead of keeping named container you can remove it after the test by replacing `--name last-kstest` option with `--rm`.
 
--- a/containers/runner/README.md
+++ b/containers/runner/README.md
@@ -67,7 +67,7 @@ Environment variables for the container (`--env` option):
 * UPDATES_IMAGE - HTTP URL of updates image to be used
 * KSTESTS_REPOSITORY - kickstart-tests git repository to be used
 * KSTESTS_BRANCH - kickstart-tests git branch to be used
-* BOOT_ISO - name of the installer boot iso from ${VOLUME_DIR}/images to be tested (default is "boot.iso")
+* BOOT_ISO - name of the installer boot iso from `${VOLUME_DIR}/images` to be tested (default is "boot.iso")
 
 
 Troubleshooting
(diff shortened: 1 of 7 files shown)


martinpitt:
I added another commit to automatically build and publish the runner container to quay.io. I manually ran the action on my [own fork](https://github.com/martinpitt/kickstart-tests/runs/1321570491?check_suite_focus=true) and used my own quay.io credentials uploaded to my fork. This resulted in a [successful container upload](https://quay.io/repository/rhn_engineering_mpitt/kstest-runner?tab=tags).

I adjusted my [Travis PoC](https://github.com/martinpitt/kickstart-tests/tree/travis) to use that container, and it [successfully ran the tests](https://travis-ci.com/github/martinpitt/kickstart-tests/builds/194752684). (The job failed because of a bug in the container runner with multiple tests, I'll look at that next).

Once merged, the rhinstaller org credentials take over, and it ought to be pubhlished under the quay.io/rhinstaller umbrella just like the [anaconda-ci](https://quay.io/repository/rhinstaller/anaconda-ci?tab=tags) container.

martinpitt:
> The job failed because of a bug in the container runner with multiple tests, I'll look at that next

Fixed and added to this PR. I [refreshed the container](https://github.com/martinpitt/kickstart-tests/runs/1322027279?check_suite_focus=true) , and re-ran Travis, and [it succeeded](https://travis-ci.com/github/martinpitt/kickstart-tests/builds/194816146).

I guess that makes this the first self-gated PR :tada:

martinpitt:
I added another bug fix to avoid clobbering the host's kickstart-tests checkout. I [refreshed the container](https://quay.io/repository/rhn_engineering_mpitt/kstest-runner?tab=tags) and [validated it on Travis](https://travis-ci.com/github/martinpitt/kickstart-tests/builds/195386762).

Note: The instructions here work, but invoking the container is not as simple as it should be. I have another commit pending that will [turn it into a simple one-liner](https://github.com/martinpitt/kickstart-tests/blob/travis/containers/runner/launch), but that requires official container builds on quay.io first. So I'll send this as a follow-up.

containers/runner/README.md
```diff
@@ -44,7 +49,7 @@ curl -L https://download.fedoraproject.org/pub/fedora/linux/development/rawhide/
 
 Run the test:
 ```
-podman run --env KSTESTS_TEST=keyboard -v ${VOLUME_DIR}:/opt/kstest/data:z --name last-kstest --device=/dev/kvm kstest-runner
+podman run --env KSTESTS_TEST=keyboard -v ${VOLUME_DIR}:/opt/kstest/data:z --name last-kstest --device=/dev/kvm rhinstaller/kstest-runner
```
jkonecny12, line 52: Should it be with the ``:latest`` tag here?

jkonecny12 (approved):
If this is really working than it a great improvement! Thanks a lot.

Especially for running VM from RAM ;).

containers/runner/README.md
martinpitt: :latest is the default if you don't specify it. As this is a command that the user has to type/run, I thought I'd aim for brevity. Unlike anaconda-ci, we will only work with `:latest` for this container (unless you explicitly want to try an earlier one). So if you prefer, I'm ok with adding the `:latest` explicitly, I'm just not sure if that was a question or a request?

martinpitt:
Hmm, the action worked fine on my fork, and the same thing seems to work fine on the anaconda project, but running the action [on master](https://github.com/rhinstaller/kickstart-tests/runs/1330424357?check_suite_focus=true) with the org credentials failed.

@jkonecny12 : What's confusing is that [anaconda](https://github.com/rhinstaller/anaconda/settings/secrets) and [kickstart-tests](https://github.com/rhinstaller/kickstart-tests/settings/secrets) have their own individual quay secrets, but the [organization](https://github.com/organizations/rhinstaller/settings/secrets) has them as well. It's enough to have them on the org, and not triplicate them. Is it ok to just remove them?

martinpitt:
Oh, of course -- `$QUAY_USERNAME` is presumably *not* just `rhinstaller`. I see that anaconda uses it only for logging in, but not for the tag name -- instead it hardcodes `rhinstaller` there. That makes the workflow harder to test on forks, I see to replace it with `github.repository_owner`.

martinpitt:
Fixed in #387.

PR #387, opened 2020-10-30
workflows: Fix runner container name on quay.io

martinpitt:
`secrets.QUAY_USERNAME` is *not* just `rhinstaller`, so hardcode the
Quay organization name like Anaconda's container workflow does.


martinpitt:
This fixes the [autobuild](https://github.com/rhinstaller/kickstart-tests/runs/1330424357?check_suite_focus=true), this was a thinko in PR #386. I thought I'd make the org name variable to be able to test this from forks, but it doesn't work like that.

jkonecny12 (approved):
LGTM!

PR #388, opened 2020-10-30
Add runner container launch script, add Travis integration

martinpitt:
This significantly simplifies the necessary setup steps to just one command. From a fresh checkout you can run
    
    containers/runner/launch keyboard [... other tests]
    
It has sensible defaults, and works both as user and through sudo (for user podman or and system podman/docker containers).

Travis now offers /dev/kvm, and its machines are powerful enough to run at least a handful of tests. This is useful for PRs that change a few tests only, to self-validate them. We can't run all tests on Travis, it would take too long. But at least for gating PRs that add/change tests, it is enough to only run these.

Travis is most likely *not* the final place where we run these, but rather we'd adjust our upshift nodes (or perhaps TMT or whatever). But it's a very cheap and easy (for us) way to get *some* initial gating into kickstart-tests. It's easy enough to remove again in the future if we want to.

### DIFF
--- a/containers/runner/README.md
+++ b/containers/runner/README.md
@@ -20,61 +20,43 @@ There is a [playbook](runner-host.yml) for deployment of the host on Fedora Clou
 
 Run a test in a container
 -------------------------
-
-Download the [automatically built](.github/workflows/container-autoupdate.yml) official container image:
+Use the [launch](./launch) script to run a set of tests from the current kickstart-tests directory in the runner container:
 ```
-podman pull quay.io/rhinstaller/kstest-runner
+containers/runner/launch keyboard [test2 test3 ...]
 ```
 
-Or build the container yourself:
+This will download the [automatically built](.github/workflows/container-autoupdate.yml) [official container image](https://quay.io/repository/rhinstaller/kstest-runner).
+
+You can also build the container yourself to test modifications to it:
 ```
 podman build -t rhinstaller/kstest-runner .
 ```
 
-Define the directory for passing of data between the container and the system (via volume):
-```
-export VOLUME_DIR=${PWD}/data
-```
+The `launch` script creates a `./data/` directory for passing of data between the container and the system (via volume). By default it downloads the current Fedora Rawhide boot.iso, but to test some other image you can put it into `data/images/boot.iso` before running `launch`.
 
-Create subdirs for test inputs (installation image) and outputs (logs):
+The result logs get written into `./data/logs/`:
 ```
-mkdir -p ${VOLUME_DIR}/images
-mkdir -p -m 777 ${VOLUME_DIR}/logs
+tree -L 3 data/logs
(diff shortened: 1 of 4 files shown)


martinpitt:
Travis results won't appear in the PR until .travis.yml lands in master (same as with github workflow checks). The run for my repo push is at https://travis-ci.com/github/martinpitt/kickstart-tests/builds/196355893 , and it succeeded.

martinpitt:
Pushed a tiny fix for SELinux. [Travis succeeded again](https://travis-ci.com/github/martinpitt/kickstart-tests/builds/197131956).

containers/runner/README.md
```diff
@@ -20,61 +20,43 @@ There is a [playbook](runner-host.yml) for deployment of the host on Fedora Clou
...
 ```
 
-Define the directory for passing of data between the container and the system (via volume):
-```
-export VOLUME_DIR=${PWD}/data
-```
+The `launch` script creates a `./data/` directory for passing of data between the container and the system (via volume). By default it downloads the current Fedora Rawhide boot.iso, but to test some other image you can put it into `data/images/boot.iso` before running `launch`.
 
-Create subdirs for test inputs (installation image) and outputs (logs):
+The result logss get written into `./data/logs/`:
```
poncovka, line 37: `The result logs ...`

scripts/run_travis.sh
```diff
@@ -0,0 +1,37 @@
...
+TESTS=$(git diff --name-only upstream/master..HEAD -- *.ks.in *.sh | sed 's/\.ks\.in$//; s/\.sh$//' | sort -u)
+
+# if the PR changes anything in the test runner, or does not touch any tests, pick a few representative tests
+if [ -z "$TESTS" ] || [ -n "$(git diff --name-only upstream/master..HEAD -- containers scripts)" ]; then
+    TESTS="$TESTS
+bindtomac-network-device-default-httpks
+container
+lvm-1
+network-device-mac-httpks
+network-static
```
poncovka, line 19: We have a test group called `coverage`, but there are about thirty tests at this moment. Maybe we could create a new group called `minimal` that would provide a reasonable subset for Travis.

poncovka (approved):
Thanks you!

containers/runner/README.md
martinpitt: No, this is parseltongue, totally correct!

:stuck_out_tongue_winking_eye: thanks for spotting, fixed.

scripts/run_travis.sh
martinpitt: I added a FIXME comment as a reminder.

containers/runner/README.md
poncovka: :snake: :snake: :snake:

rvykydal (approved):
Looks great to me.

PR #389, opened 2020-11-02
Move kstests-in-cloud.sh into scripts/

martinpitt:
This avoids run_kickstart_tests.sh's find_tests() to consider it as
test, and avoids this error:

```
make: Entering directory '/opt/kstest/kickstart-tests/scripts'
make: *** No rule to make target 'COMMAND'.  Stop.
```

his is a bit difficult to avoid, as we do want to keep it
executable, and find_tests() can't be changed to look for `*.ks.in`
instead as that would break setting `$KICKSTART_NAME` in a test (even
though none currently use it)

We could also rename it to drop the `.sh` suffix, but this seems to fit
better in the scripts/ directory anyway.


martinpitt:
@rvykydal : Filing this separately, as this has the potential to break the Jenkins tests -- I fixed all references within this repo, and it doesn't seem to be referenced in any Ansible playbooks -- so it's very likely that this gets called from some Jenkins config that isn't visible from outside.

If this is too intrusive, we can also fix this in a different way, like putting an extra `grep -q TESTTYPE=` into find_tests() or something such.

rvykydal (approved):
Looks good to me, thank you.

VladimirSlavik (approved):
Looks good to me, thank you!

PR #390, opened 2020-11-02
Various small runner container fixes

martinpitt:
Except for a tiny change in run_kickstart_tests.sh, all of these only affect the runner container, which we don't use in production yet -- so this should be safe to land.


rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):
Looks good to me.

PR #391, opened 2020-11-02
Support running all tests in runner container

martinpitt:


rvykydal (approved):
Looks good to me.
Originally I required tests to be run to be specified explicitly because I was afraid of running all tests (requiring a lot of resources) in the default case (or a case "I don't really know what I am doing").

martinpitt:
> Originally I required tests to be run to be specified explicitly because I was afraid of running all tests

That's a good point. How about we only do that when calling `launch all`? That would mean we could never have an actual test `all.sh`, but that may be an acceptable trade-off.

martinpitt:
@rvykydal: Reworked to explicitly use `launch all`.

rvykydal (approved):
Looks good to me, thank you.

PR #392, opened 2020-11-02
Support local HTTP caching for running tests

martinpitt:
This greatly speeds up tests locally and avoids bandwidth bottlenecks when running tests in parallel.

Also enable squid in Travis. As that has huge bandwidth, it does not help too much with the test runtime, but it validates the caching implementation.

Caching https isn't possible, and the cert hacks for explicit `CONNECT` are too intrusive for proper testing, so this requires running most tests against a http:// source (and relying on GPG signature verification). I added a single https test to make sure that https still works as well.

I'm aware that this may be controversial -- it's a trade-off. If you don't like this as-is, I can also look into making the default repo more dynamic, and provide a knob to the test to switch to plain http by default. I appreciate some guidance here.

### DIFF
--- a/https-repo.sh
+++ b/https-repo.sh
@@ -1,0 +1,22 @@
+#
+# Copyright (C) 2020  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+# Red Hat Author(s): Martin Pitt <mpitt@redhat.com>
+
+TESTTYPE="method"
+
+. ${KSTESTDIR}/functions.sh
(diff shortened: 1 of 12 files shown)


martinpitt:
Oh, two things: The nightly tests override the default repo to http://download.eng.brq.redhat.com/, so (1) they already use plain http, and (2) the change here doesn't affect them either way. However, it's very likely that the new `https-ks` test fails there. @rvykydal , how exactly does the infra change the default repo? I'd like to reproduce this locally, and adjust the test accordingly.

rvykydal:
The infra adds/replaces variable is scripts/defaults.sh on runner checkout of kicstart-tests repository by the test run script:
https://github.com/rvykydal/kickstart-tests/blob/10a1123b157f2cca569a56a27f2bed623a50c53f/ansible/roles/kstest-master/templates/run_tests.sh.j2#L107
based on a test run configuration:
https://github.com/rvykydal/kickstart-tests/blob/10a1123b157f2cca569a56a27f2bed623a50c53f/ansible/roles/kstest-master/defaults/main/test-configuration.yml#L50

martinpitt:
Thanks @rvykydal : I modified the anaconda install locally to use the eng.brq mirror, and the test case still succeeds. The *configured* yum sources in /etc/yum.repos.d/ are still using fedoraproject.org, and are apparently not set to the mirror that anaconda itself uses.

https-repo.sh
```diff
@@ -0,0 +1,29 @@
...
+# Red Hat Author(s): Martin Pitt <mpitt@redhat.com>
+
+TESTTYPE="method"
+
+. ${KSTESTDIR}/functions.sh
+
+prepare() {
+    ks=$1
+    tmpdir=$2
+    sed -e 's|http://\(.*fedoraproject.org\)|https://\1|' ${ks} > ${tmpdir}/kickstart.ks
```
poncovka, line 27: I would prefer to include a different fragment in the kickstart file. Please, create a new file `fragments/platform/fedora_rawhide/repos/secure.ks` with the old content of `default.ks`. Then you can replace `%ksappend repos/default.ks` with `%ksappend repos/secure.ks`.

martinpitt: Good idea, this is less hacky. Done, PTAL?

poncovka (approved):
It looks great. Thanks!

PR #393, opened 2020-11-02
linchpin: fix getting of anaconda version in a test run

rvykydal:
The installer image used to be in /LiveOS/rootfs.img of install.img squashfs
image, but recently in Fedora it has changed to be directly the content of
install.img.


jkonecny12 (approved):
Looks good to me I guess. I'm starting to be a bit lost here :).

VladimirSlavik (approved):
I never found myself :P I think it should do what Radek says? I understand that as: Check if there is a rootfs file; if so, mount that to peel away one more layer of the onion; if not, the layer is not there so stuff can be accessed directly already from the last mount.

PR #394, opened 2020-11-03
Fix permissions of some tests

martinpitt:
run_kickstart_tests.sh's `find_test()` depends on tests being
executable.

The two unified* ones are knownfailures, but let's leave the filtering
to that testtype.


martinpitt:
In my [local run](https://github.com/rhinstaller/kickstart-tests/pull/373#issuecomment-708208161) I get a failure in snapshot-post:

    dbus.exceptions.DBusException: ('com.redhat.lvmdbus1.Vg', 'Exit code 5, stderr =   Logical volume testvg/testLV contains a filesystem in use.\n  Can\'t deactivate volume group "testvg" with 1 open logical volume(s)\n')

This doesn't look like a runner container specific problem, but let's run these on the official infra to compare.

Note that the Travis test here will/should fail, as it also runs the knownfailures. But I still think it's the right thing to do.

rvykydal:
One thing to note here is that KNOWNFAILURE tests are currently filtered out by run_kickstart_tests.sh (and our tooling wrapping it) only when no tests are specified (ie all tests are run). If I run for example tests of type "network", knownfailure tests are run by default. It can be of course updated by skipping tests of type KNOWNFAILURE explicitly so perhaps this note is a non-issue.

martinpitt:
Indeed, the tests on travis faithfully reproduce the knownfailures and what I mentioned with snapshot-pre. However, this won't break other Travis tests, as this is only running the "affected" ones.

martinpitt:
@rvykydal : Yes, but in general the idea of "run affected tests" is that when a PR touches a test it's (hopefully) because it wants to fix it. Not running them at all in Travis wouldn't make much sense IMHO, as long as they are PR specific and not part of the "default" set.

rvykydal (approved):

jkonecny12 (approved):
LGTM!

PR #396, opened 2020-11-03
Two runner container fixes for highly parallel runs

martinpitt:


martinpitt:
With these two fixes I just ran a massive "run all tests" with `$TEST_JOBS=25` on one of our e2e machines (32 CPUs, 196 GiB RAM). I can't run more because I have neither enough RAM nor enough disk available for the sizable /var/tmp :grin: 

But even with 25 it looks good. I got two failures aside from the ones known in #373: bindtomac-bridged-bond-httpks and onboot-activate-httpks. I'll look at the details tomorrow. But these two fixes seem good.

martinpitt:
I filed https://bugzilla.redhat.com/show_bug.cgi?id=1894359 about the virt-install storage pool race; let's see if upstream has some ideas how to debug/fix that.

martinpitt:
The commit message was wrong, it still said "3 GB". Fixed that, no code changes.

jkonecny12 (approved):
Wow, pretty impressive changes. Thanks!

rvykydal (approved):
Looks good to me.

PR #397, opened 2020-11-03
Ignore "Call Trace:" in logs temporarily

rvykydal:
Ignore Call Traces so that rhbz#1885978 and rhbz#1893950 don't block Anaconda
testing and development of testing tools.

### DIFF
--- a/scripts/launcher/lib/log_monitor/log_handler.py
+++ b/scripts/launcher/lib/log_monitor/log_handler.py
@@ -31,7 +31,8 @@ class VirtualLogRequestHandler(LogRequestHandler):
         "CRIT mdadm:DegradedArray event detected",
 
         # Ignore a call trace during debugging.
-        # "Call Trace:"
+        # Enabled tempoarily for rhbz#1885978, rhbz#1893950
+        "Call Trace:"
     ]
 
     # Specify error lines you want to add on top


jkonecny12 (approved):
You made me a sad panda but still you have my ACK.
Even thought it would be great to describe the reason in commit message ideally with the part of the Call trace?

rvykydal:
> Even thought it would be great to describe the reason in commit message ideally with the part of the Call trace?
There are links to BZs, where anyone can find the call trace.

poncovka (approved):
Thank you!

martinpitt:
Obvious question, but could we ignore particular traces only? so that we can associate them to bugzillas, but still find new ones?

rvykydal:
> Obvious question, but could we ignore particular traces only? so that we can associate them to bugzillas, but still find new ones?

I've looked into code to double-check but the context for detection of the failure is single line, which is
```
01:40:57,030 WARNING kernel:Call Trace:
```
so I think with current mechanism it is not possible.
If we wanted to detect error on more specific message, like
```
01:40:57,018 WARNING kernel:WARNING: possible circular locking dependency detected
```
it could be easily added. But we want to ignore Call Trace in this case which would require enhancing of the mechanism.

PR #400, opened 2020-11-05
Fix the kickstart test of the default filesystem type on Fedora Server

poncovka:
Let's use the same approach as in https://github.com/rhinstaller/kickstart-tests/pull/395.


martinpitt (approved):
Dekuji!

PR #402, opened 2020-11-09
Support downloading and testing against rawhide+COPR iso

martinpitt:
PR #403 introduced a GitHub workflow for building a boot.iso from Fedora Rawhide and the anaconda/blivet/dnf COPRs. This provides an easy external regression test for these projects before they land in Fedora.

The boot.iso (plus some other files) are exported as an artifact.
Downloading them requires authentication, require a GitHub token file as
value of the new `--daily-copr` option. When given, download/unpack that
instead of the official Rawhide boot.iso.

---

This is mostly a PoC for now, to demonstrate that this can provide a full end-to-end workflow (without requiring any internal infrastructure). The `launch` script may not be the final place for this, but it's easy enough to move the code someplace else.

I tested this with my personal GitHub access token and also with the "rhel8-runner" one in our internal secrets repository.

 - [x] Land PR #402
 - [x] Adjust the repository API URL afterwards
 - [x] Land PR #404 
 - [x] Turn env var into a CLI option

### DIFF
--- a/containers/runner/launch
+++ b/containers/runner/launch
@@ -29,18 +29,22 @@ Options:
  -j, --jobs N                         Run N jobs in parallel (default: $(nproc))
  -t, --testtype TYPE                  Only run TYPE tests
  -s, --skip-testtypes TYPE[,TYPE..]   Don't run tests with given types
+ --daily-iso TOKEN_FILE               Download and use daily boot.iso instead of rawhide's
+                                      (This requires a GitHub token that can read
+                                       rhinstaller/kickstart-tests workflow artifacts.)
  -h, --help                           Show this help
 EOF
 }
 
 # parse options
-eval set -- "$(getopt -o j:t:s:h --long jobs:,testtype:,skip-testtypes:,help -- "$@")"
+eval set -- "$(getopt -o j:t:s:h --long jobs:,testtype:,skip-testtypes:,daily-iso:,help -- "$@")"
 
 while true; do
     case "${1:-}" in
         -j|--jobs) shift; TEST_JOBS=$1 ;;
         -t|--testtype) shift; TESTTYPE=$1 ;;
         -s|--skip-testtypes) shift; SKIP_TESTTYPES=$1 ;;
+        --daily-iso) shift; DAILY_ISO_TOKEN="$1" ;;
         -h|--help) usage; exit 0 ;;
         --) shift; break ;;
     esac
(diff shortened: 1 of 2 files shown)


jkonecny12 (approved):
Looks great to me. Let's see how this will work when everything is ready!

martinpitt:
PR #402 landed, I [triggered the build manually](https://github.com/rhinstaller/kickstart-tests/runs/1390405571), so we have a first "official" boot.iso to play with. I updated this PR to adjust the repository location.

I still would like to use a proper `--daily-copr-iso TOKEN` for that instead of the much less discoverable env var. Hence I'll block this on PR #404, and once that lands I'll adjust the CLI accordingly.

martinpitt:
I reworked this a little bit:
 * Turned into a `--daily-copr TOKEN` option with help, which is a little more tasteful than the magic env var
 * Documented in README
 * Robustified the zip download (tested what happens with absent,  broken, and insufficiently privileged token)
 * Added fallback to p7zip, as `unzip` is not available on RHEL 7 and I'd like to run this on our e2e machines.

martinpitt:
#404 landed, rebased, ready for review now.

containers/runner/launch
```diff
@@ -29,18 +29,22 @@ Options:
  -j, --jobs N                         Run N jobs in parallel (default: $(nproc))
  -t, --testtype TYPE                  Only run TYPE tests
  -s, --skip-testtypes TYPE[,TYPE..]   Don't run tests with given types
+ --daily-copr TOKEN_FILE              Download and use daily COPR boot.iso instead of rawhide's
```
jkonecny12, line 32: Not sure if we want to named it ``copr``. It seems to me like an implementation detail. Maybe having more generic name ``daily-build-iso`` would be better?

containers/runner/launch
```diff
@@ -61,8 +65,19 @@ fi
 mkdir -p data/images
 mkdir -p -m 777 data/logs
 if ! [ -e data/images/boot.iso ]; then
-    echo "INFO: data/images/boot.iso does not exist, downloading current Fedora Rawhide Server image..."
-    curl -L https://download.fedoraproject.org/pub/fedora/linux/development/rawhide/Server/x86_64/os/images/boot.iso --output data/images/boot.iso
+    if [ -n "${DAILY_COPR_TOKEN:-}" ]; then
+        echo "INFO: data/images/boot.iso does not exist, downloading daily rawhide+COPR iso..."
+        CURL="curl -u token:$(cat $DAILY_COPR_TOKEN) --show-error --fail"
```
jkonecny12, line 70: Hmm I wonder why we need token to download test artifacts which are present publicly...

jkonecny12 (dismissed):
Looks good to me. Just a few nitpicks/ideas?

containers/runner/launch
martinpitt: Names are always difficult! I wanted to point out that it's not just a daily Rawhide iso. But I'm fine with calling the latter "rawhide" and this one "daily".

containers/runner/launch
martinpitt: They aren't -- if you log out of GitHub (or use private mode), you can't see the artifacts. You just get a note like "log in to see the artifacts". Curiously you don't even need to have particular permissions for *this* project -- I tried this with a cockpituous token just for fun, and that worked. To me it mostly just feels like a mild prevention of abusing GitHub as a generic file upload/download place, so that you can't do anonymous illegal file stores/activities.

containers/runner/launch
martinpitt: Renamed to `--daily-iso` as per our IRC discussion.

jkonecny12 (approved):
Looks even better right now!

PR #403, opened 2020-11-10
Add workflow for building daily boot.iso

martinpitt:
This replaces the ad-hoc upshift instance "nightly_iso_builder" and the Jenkins job "daily_boot_iso". There is nothing secret going on here, this does not even need any secrets, so it's fine to let this run on public infra in GitHub.

With that, kickstart-tests running anywhere can make use of the latest daily boot.iso (even though they will need a GitHub token to be allowed to download the zip artifact) - i.e. we don't restrict access to that to just upshift any more. Developers can just as easily download/use it in a local test run.

PR #402 shows how to use it. But that may still require some discussions, so I suggest to land this job early so that we can see how it works out, and give us some isos to play with?

### DIFF
--- a/.github/workflows/daily-boot-iso.yml
+++ b/.github/workflows/daily-boot-iso.yml
@@ -1,0 +1,42 @@
+name: Build daily Rawhide+COPR boot.iso
+on:
+  schedule:
+    - cron: 0 22 * * *
+  # be able to start this action manually from a actions tab when needed
+  workflow_dispatch:
+
+jobs:
+  boot_iso:
+    name: Build boot.iso
+    runs-on: ubuntu-latest
+    steps:
+      # lorax and its dependencies do not exist in Ubuntu, and we want to test the latest Fedora version anyway
+      # lorax does mounts and uses loop devices, thus needs to be privileged
+      - name: Start fedora container
+        run: |
+          mkdir data
+          docker run --name fedora -d --privileged --network host -v $PWD/data:/data fedora:latest sleep infinity
+
+      - name: Install lorax
+        run: docker exec fedora dnf install -y lorax
+
+      - name: Run lorax
+        run: docker exec -w /data fedora lorax -p Fedora -v 34 -r 34 -s http://download.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os/ -s https://download.copr.fedorainfracloud.org/results/rpmsoftwaremanagement/dnf-nightly/fedora-rawhide-x86_64/ -s https://download.copr.fedorainfracloud.org/results/@rhinstaller/Anaconda/fedora-rawhide-x86_64/ -s https://copr-be.cloud.fedoraproject.org/results/@storage/blivet-daily/fedora-rawhide-x86_64/ /data/results
+
+      - name: Make generated files accessible
+        run: docker exec fedora chmod -R a+rX /data/
+
+      - name: Upload log artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: logs
+          path: |
+            data/*.log
+            data/*.txt
+
+      - name: Upload image artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: images
+          path: |
+            data/results/**


martinpitt:
I previously sent this to https://github.com/rhinstaller/devel-tools/pull/25 , but now that I know where that iso is consumed, I think it makes more sense here.

rvykydal (approved):
Looks good to me

.github/workflows/daily-boot-iso.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Start fedora container
+        run: |
+          mkdir data
+          docker run --name fedora -d --privileged --network host -v $PWD/data:/data fedora:latest sleep infinity
+
+      - name: Install lorax
+        run: docker exec fedora dnf install -y lorax
+
+      - name: Run lorax
+        run: docker exec -w /data fedora lorax -p Fedora -v 34 -r 34 -s http://download.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os/ -s https://download.copr.fedorainfracloud.org/results/rpmsoftwaremanagement/dnf-nightly/fedora-rawhide-x86_64/ -s https://download.copr.fedorainfracloud.org/results/@rhinstaller/Anaconda/fedora-rawhide-x86_64/ -s https://copr-be.cloud.fedoraproject.org/results/@storage/blivet-daily/fedora-rawhide-x86_64/ /data/results
```
jkonecny12, line 24: I wonder that it's that simple to call Lorax from container because in general that is pain. However, if this works than it's great!

.github/workflows/daily-boot-iso.yml
```diff
@@ -0,0 +1,42 @@
...
+          path: |
+            data/*.log
+            data/*.txt
+
+      - name: Upload image artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: images
+          path: |
+            data/results/**
```
jkonecny12, line 42: I hope there is no quota for file storage. However, let's see how this will work and we can think of other solution later if there will be a problem.

jkonecny12 (approved):
Great work @martinpitt! It's much simpler than I would expect!

.github/workflows/daily-boot-iso.yml
```diff
@@ -0,0 +1,42 @@
...
+  boot_iso:
+    name: Build boot.iso
+    runs-on: ubuntu-latest
+    steps:
+      # lorax and its dependencies do not exist in Ubuntu, and we want to test the latest Fedora version anyway
+      # lorax does mounts and uses loop devices, thus needs to be privileged
+      - name: Start fedora container
+        run: |
+          mkdir data
+          docker run --name fedora -d --privileged --network host -v $PWD/data:/data fedora:latest sleep infinity
```
jkonecny12, line 18: However, I wonder why ``sleep infinity``. I thought it should be fine to use ``docker create`` without a command, or am I missing something?

.github/workflows/daily-boot-iso.yml
martinpitt: It really is, as lorax is using loop devices and mounts. Hence the need for `--privileged` (see the comment in `steps:`).

.github/workflows/daily-boot-iso.yml
martinpitt: I thought about that, too. I set the [workflow retention period](https://github.com/rhinstaller/kickstart-tests/settings/actions) to 7 days for now. If that is somehow not respected, or we need to keep other logs longer, then we could add another workflow which regularly cleans up all boot.iso except for the most two recent ones.

.github/workflows/daily-boot-iso.yml
martinpitt: `create` doesn't start the container, thus you can't run anything in it. The container will run as long as its pid 1 is running, and thus that needs to be something innocuous. `bash -i` is an ususal suspect, but that doesn't work with `--detach`, it would just immediately exit.

.github/workflows/daily-boot-iso.yml
jkonecny12: ohh, I missed the privileged. In that case, yes, it will work :)

.github/workflows/daily-boot-iso.yml
jkonecny12: Ok, thanks for info.

.github/workflows/daily-boot-iso.yml
jkonecny12: I see. Thanks for explanation!

PR #405, opened 2020-11-12
Update defaults of linchpin tooling configuration

rvykydal:
Note: to use default.sh defaults the values are set to "" in test configuration.

### DIFF
--- a/ansible/roles/kstest-master/defaults/main/test-configuration.yml
+++ b/ansible/roles/kstest-master/defaults/main/test-configuration.yml
@@ -4,7 +4,7 @@
 ### The subject to be tested.
 
 # Installer boot iso url.
-kstest_boot_iso_url: "https://download.fedoraproject.org/pub/fedora/linux/releases/29/Everything/x86_64/os/images/boot.iso"
+kstest_boot_iso_url: "https://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os/images/boot.iso"
 
 # Url of updates image to be used.
 kstest_updates_img: ""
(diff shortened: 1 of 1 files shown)


ansible/roles/kstest-master/defaults/main/test-configuration.yml
```diff
@@ -47,10 +47,10 @@ kstest_override_repos: "yes"
 # Base repository url
 # In the form of kickstart url command option.
 #kstest_url: "--mirror=http://mirrors.fedoraproject.org/mirrorlist?repo=fedora-rawhide\\&arch=$basearch"
-kstest_url: "--url=https://download.fedoraproject.org/pub/fedora/linux/releases/29/Everything/x86_64/os"
+kstest_url: "--url=https://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os/"
 
 # Software ftp repository for tests.
-kstest_ftp_url: "ftp://mirror.utexas.edu/pub/fedora/linux/releases/29/Everything/x86_64/os/"
+kstest_ftp_url: "ftp://ftp.tu-chemnitz.de/pub/linux/fedora/linux/development/rawhide/Everything/x86_64/os/"
```
martinpitt, line 53: Why does this need to duplicate the values in scripts/defaults.sh at all? Could we just leave these empty, so that we have a single place of truth, but still the option to override it locally?

martinpitt (approved):
Thanks! One question, but surely this is an improvement.

ansible/roles/kstest-master/defaults/main/test-configuration.yml
rvykydal: Originally I wanted the tooling to have repo defaults of its own (in test-configuration.yml) independent of the defaults/scripts.sh (which were not updated much) because I wanted to provide default configuration with both boot.iso url and corresponding repo urls for user, and make the repo configuration required in a way - making sure repos corresponding to the boot.iso are used, not repos from defaults.sh. (Because although it is often possible to install with different versions of installer and repository it is not supported in general)

But given how the tooling is (not) used I am OK with not setting the values.

PR #406, opened 2020-11-12
Simplify daily-boot-iso workflow

martinpitt:
It's possible to specify options (in particular, `--privileged`) for the launched container. So use the declarative syntax instead of launching the container manually, and entirely avoid the /data volume. [successful run on my fork](https://github.com/martinpitt/kickstart-tests/actions/runs/360185931)

The second commit greatly reduces the size of the artifact, see commit message for details. [successful run on my fork](https://github.com/martinpitt/kickstart-tests/actions/runs/360231363)

### DIFF
--- a/.github/workflows/daily-boot-iso.yml
+++ b/.github/workflows/daily-boot-iso.yml
@@ -9,34 +9,28 @@ jobs:
   boot_iso:
     name: Build boot.iso
     runs-on: ubuntu-latest
-    steps:
-      # lorax and its dependencies do not exist in Ubuntu, and we want to test the latest Fedora version anyway
+    container:
+      image: fedora:latest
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
Great improvement! Thanks!

jkonecny12:
BTW  may I disable boot iso creation on Jenkins? @rvykydal @martinpitt

martinpitt:
@jkonecny12 : Not quite yet -- this is all just affecting the new-style container workflow, and at least  PR #402 still needs to land for that. But the current infra doesn't know about any of that yet. That's the topic of next Wednesday's meeting.

martinpitt:
I land this now, to reduce the artifact size on github. Easy enough to adjust later if @rvykydal has some input how to change this.

PR #407, opened 2020-11-16
Fix squid container proxy for firewalld and SELinux

martinpitt:
With these patches I can successfully run kickstart-tests with the local http proxy on our RHEL 7 e2e machines:
```
# containers/squid.sh start
# containers/runner/launch --skip-testtypes rhel-only,knownfailure --jobs 27 all
```
Last time I tried that was with 25 parallel jobs and no proxy, which took about 2:54 hours. (The machine does not have enough RAM/storage to run 32 tests in parallel).

This time it took just 2:20 hours, so a noticeable speedup. Of these 34 minutes, 14 can be explained due to the higher parallelism, and 20 due to the download caching.

### DIFF
--- a/containers/squid.sh
+++ b/containers/squid.sh
@@ -14,7 +14,7 @@ CRUN=${CRUN:-$(which podman docker 2>/dev/null | head -n1)}
 if [ "${1:-}" = start ]; then
     # This image is well-maintained (auto-built) and really small
     $CRUN run --net host --name squid --detach \
-        --volume "$MYDIR"/squid-cache.conf:/etc/squid/conf.d.tail/cache.conf:ro \
+        --volume "$MYDIR"/squid-cache.conf:/etc/squid/conf.d.tail/cache.conf:ro,z \
         --volume ks-squid-cache:/var/cache/squid docker.io/b4tman/squid
 
     # Redirect all traffic from external interfaces (like container bridges) through our local proxy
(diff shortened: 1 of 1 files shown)


jkonecny12 (changes requested):
Looks good to me. However, do we want to support both ``iptables`` and ``nft``? Wouldn't it be easier to just use iptables for everything?

Also do we need to support RHEL-7? Anaconda won't be released on RHEL-7 anymore so there is no reason to keep testing it.

martinpitt:
@jkonecny12 : iptables is being deprecated, and Eric Garver says that nft is the new way forward. On developer machines we should use nft, like firewalld itself.

About RHEL 7, this isn't about testing a RHEL-7 install, but running the runner container *on* a RHEL 7 system. Unfortunately our e2e machines are stuck on RHEL 7 and can't currently be reinstalled with RHEL 8. We'll discuss that on Wednesday, but for now I think these are our best bet.

martinpitt:
@jkonecny12 : I took out the iptables patch. I can put it back in if/when we need to run on e2e, or perhaps Satellite gets fixed by then :crossed_fingers:

jkonecny12:
> @jkonecny12 : I took out the iptables patch. I can put it back in if/when we need to run on e2e, or perhaps Satellite gets fixed by then crossed_fingers

Sounds great to me. Thanks!

jkonecny12 (approved):
LGTM!

PR #408, opened 2020-11-18
Change Fedora mirror in daily boot.iso build

martinpitt:
The geo-location mirror often picks download-ib01.f.o. which is very
unreliable. Use another mirror close to the Azure cloud.

----

The [last 4 runs](https://github.com/rhinstaller/kickstart-tests/actions?query=workflow%3A%22Build+daily+Rawhide%2BCOPR+boot.iso%22) failed on random mirror download errors, which is annoying. I [did a run on my fork with this mirror](https://github.com/martinpitt/kickstart-tests/runs/1416544606?check_suite_focus=true), and it went well. It downloaded the entire package set in just 20s (!) Let's try this for a bit?

### DIFF
--- a/.github/workflows/daily-boot-iso.yml
+++ b/.github/workflows/daily-boot-iso.yml
@@ -18,7 +18,8 @@ jobs:
         run: dnf install -y lorax
 
       - name: Run lorax
-        run: lorax -p Fedora -v 34 -r 34 -s http://download.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os/ -s https://download.copr.fedorainfracloud.org/results/rpmsoftwaremanagement/dnf-nightly/fedora-rawhide-x86_64/ -s https://download.copr.fedorainfracloud.org/results/@rhinstaller/Anaconda/fedora-rawhide-x86_64/ -s https://copr-be.cloud.fedoraproject.org/results/@storage/blivet-daily/fedora-rawhide-x86_64/ /tmp/results
+        # The download.fedoraproject.org automatic redirector often selects download-ib01.f.o. for GitHub's cloud, which is too unreliable; manually select a good one that is nearby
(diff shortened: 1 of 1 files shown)


.github/workflows/daily-boot-iso.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         run: dnf install -y lorax
 
       - name: Run lorax
-        run: lorax -p Fedora -v 34 -r 34 -s http://download.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os/ -s https://download.copr.fedorainfracloud.org/results/rpmsoftwaremanagement/dnf-nightly/fedora-rawhide-x86_64/ -s https://download.copr.fedorainfracloud.org/results/@rhinstaller/Anaconda/fedora-rawhide-x86_64/ -s https://copr-be.cloud.fedoraproject.org/results/@storage/blivet-daily/fedora-rawhide-x86_64/ /tmp/results
+        run: lorax -p Fedora -v 34 -r 34 -s http://pubmirror2.math.uh.edu/fedora-buffet/fedora/linux/development/rawhide/Everything/x86_64/os/ -s https://download.copr.fedorainfracloud.org/results/rpmsoftwaremanagement/dnf-nightly/fedora-rawhide-x86_64/ -s https://download.copr.fedorainfracloud.org/results/@rhinstaller/Anaconda/fedora-rawhide-x86_64/ -s https://copr-be.cloud.fedoraproject.org/results/@storage/blivet-daily/fedora-rawhide-x86_64/ /tmp/results
```
jkonecny12, line 21: Please add a comment explaining why we don't have mirror list here.

jkonecny12 (approved):
Other than that it's fine!

rvykydal (dismissed):
Looks good to me.

.github/workflows/daily-boot-iso.yml
martinpitt: *Can* we give it a mirror list? If that works, then that'd be much better of course, then dnf would hopefully try a different mirror. I wasn't aware that lorax supports that (and googling didn't really help). I'll try that then.

martinpitt: jkonecny | pitti, ups, my fault! I did not meant mirrorlist but that proxy or load-balancer

martinpitt: Added a comment.

jkonecny12 (approved):
Looks great to me, now!

martinpitt:
I [triggered a manual run](https://github.com/rhinstaller/kickstart-tests/actions/runs/372685758) and that succeeded. Let's see how it holds up the next few nights.

PR #409, opened 2020-11-18
Support testing local updates.img file, support updates image in runner container launch script

martinpitt:
This makes the validation of local anaconda changes really easy for a developer or CI.

I tested this without an updates.img (Travis does that as well), with an URL, and with a local file:

    containers/runner/launch ../anaconda/updates.img keyboard


jkonecny12 (approved):
I've tested this and it works pretty nice. Even thought my test timed out but it doesn't seem to be problem here.

PR #411, opened 2020-11-19
Pull in libvirt pool/volume TOCTOU fix

martinpitt:
This often breaks our tests with non-trivial parallelism. Grab the fix
from GitHub as a patch, backport it to the older 2.2.1 version in Fedora
32, and hot-apply it in the container.

This will (and is supposed to) fail once a new libvirt version lands in
Fedora 32 with the fix. We will notice that in a failed
container-autoupdate.yml workflow, when that happens we can drop this
hack.

### DIFF
--- a/containers/runner/Dockerfile
+++ b/containers/runner/Dockerfile
@@ -20,6 +20,11 @@ RUN dnf -y update && \
     openssh-clients && \
     dnf -y clean all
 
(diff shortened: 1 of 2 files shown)


martinpitt:
I built a local container with this and successfully ran tests in it.

rvykydal (approved):
Looks good to me.

poncovka (approved):
Interesting solution :-) Thanks!

PR #413, opened 2020-11-19
Fix unpacking of daily boot.iso artifact

martinpitt:
Commit 929b3f1617ea2 changed the artifact to only contain boot.iso. That
had the side effect that the .zip does not contain an images/
subdirectory any more. Adjust the unpacking to follow suit -- it's
actually nicer that way to not make assumptions about the zip's
structure.

----

I noticed that bug while trying to set up the daily boot.iso run on my fork. I included this commit into my fork, and the test  [happily runs now](https://github.com/martinpitt/kickstart-tests/runs/1424649220?check_suite_focus=true) -- at least it's past the "unpack zip" stage that failed before.


rvykydal (approved):
Looks good to me.

PR #414, opened 2020-11-20
Support KSTEST_* substitution in fragments

martinpitt:
Avoid duplicating the default Fedora rawhide repositories in the
default.ks fragment and scripts/default.sh. This makes it much easier to
set a local mirror in a test environment.

To implement this, run the %ksappend substitution before the
`@KSTEST_*@` sedding, so that the sed will catch the substitutions in
the fragments.

To make that work nicely, simplify apply-ksappend.py to only process one
file at a time (specified as positional command line argument, as it's
not really optional and thus should not be an option), avoid all the
file copying, and write the result to stdout, so that it can be
pipelined efficiently.

Also make any errors in apply-ksappend.py fatal. We don't want to hide
substitution errors or missing files in our tests, let's just fix them
instead.

Replace the ArgumentParser class in apply-ksappend.py with a simple
function. The class did not add any new functionality, and was just
unnecessary glue.

### DIFF
--- a/scripts/run_kickstart_tests.sh
+++ b/scripts/run_kickstart_tests.sh
@@ -286,7 +286,6 @@ export KSTESTDIR=$(pwd)
 # The name of input kickstart (ks.in) file is
 # 1) either defined in the test (.sh) file by KICKSTART_NAME variable
 # 2) or the same as the test (.sh) file name .sh if the variable is not found
-tests_ks=""
 for t in ${tests}; do
     ksname_line=$(grep KICKSTART_NAME= ${t})
     if [[ -n "$ksname_line" ]]; then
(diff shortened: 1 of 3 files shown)


scripts/run_kickstart_tests.sh
```diff
@@ -296,11 +296,10 @@ for t in ${tests}; do
         ks=${t/.sh/.ks.in}
     fi
     tests_ks+="${ks/.ks.in/.ks} "
```
rvykydal, line 298: Can we get rid of tests_ks completely? I don't see it used elsewhere.

rvykydal (approved):
Looks good to me but I'd wait for @M4rtinK's eye.

scripts/run_kickstart_tests.sh
martinpitt: Good point, thanks! I'll do that, but I'll first let @M4rtinK review this in case he has more fundamental objections.

martinpitt: Done (it was quick, and with that we already get a new test run)

M4rtinK (approved):
Looks good to me - nice improvements & cleanups! :-)

IIRC it was not really clear what will and what will not be handled by the kickstart fragments back when the substitution scrip was written & we never got back to it clean it up & adjust it to the actual prevailing kickstart fragment usage. It'ts really nice to see it being done now - thanks a lot! :)

PR #412, opened 2020-11-19
Run daily kickstart test scenarios in GitHub workflow

martinpitt:
[Latest run on my fork for rawhide+daily](https://github.com/martinpitt/kickstart-tests/actions/runs/382599422) - just one random failure

Older runs:
 - [rawhide+minimal](https://github.com/martinpitt/kickstart-tests/actions/runs/371938555), [second run](https://github.com/martinpitt/kickstart-tests/actions/runs/378675867)
 - [rawhide-text + minimal](https://github.com/martinpitt/kickstart-tests/actions/runs/378769793)
 - ~[old daily+minimal](https://github.com/martinpitt/kickstart-tests/actions/runs/372434207) - that  run looks rather sorry, need to investigate more closely; the networking failures reproduce and are consistent, and most likely an actual regression on anaconda master; the timeouts looks like a regression in dnf copr? they happen on the official results as well.~
- [current daily boot.iso](https://github.com/martinpitt/kickstart-tests/runs/1447171006?check_suite_focus=true) -- tests succeeded, but the script failed overall -- fixed
 - [rawhide + r-text + daily](https://github.com/martinpitt/kickstart-tests/actions/runs/380457486)
 - [rawhide + daily](https://github.com/martinpitt/kickstart-tests/actions/runs/381259209) -- first daily success, rawhide timed out

 - [x] Builds on top of PR  #410 
 - [x] Requires fix in PR #411 and container refresh to  make tests run reliably
 - [x] daily boot.iso scenario requires fix in PR #413
 - [x] add option to override defaults.sh and kernel boot args: PR #415 
 - [x] Infra for creating the workers (builder PR 30)
 - [x] Restrict daily-iso scenario to `packaging` tests only, to mirror what the current tests do and make the scenario more reliable (it's mostly for the benefit of the dnf team)
 - [x] Fix [daily-iso test failure](https://github.com/martinpitt/kickstart-tests/runs/1447171006?check_suite_focus=true) (all tests passed) -- due to nfs-repo-and-addon failure
 - [x] Mark all long-standing failures as `knownfailure` for now, so that tests have a chance to get green: PR #418
 - [ ] ~Fix `network-noipv4-pre` failure, this doesn't happen on cobra02~ **update**: this is a race condition, sometimes it succeeds; depends on picking ens4 vs. enp1s0 style network names
[...]

PR #415, opened 2020-11-23
More configuration options for launch script

martinpitt:
Also slightly simplify handling an undefined `$VAR_TMP`.

----

We use this for running tests in upshift (PR #412)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -221,6 +221,10 @@ The following environment variables are currently supported:
   KSTEST_OS_VERSION = 26, Fedora rawhide has "Rawhide", and RHEL 7.3 has
   KSTEST_OS_VERSION = 7.3 .
 
+- KSTEST_EXTRA_BOOTOPTS - This variable is used in functions.sh to pass
+  additional kernel command line options. For example, setting this to `inst.text`
+  enables Anaconda's text mode (instead of the default GUI).
+
 Chapter 3. Sharing common code in kickstart (.ks.in) files
 ==========================================================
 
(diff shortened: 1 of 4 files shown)


martinpitt:
I added two more commits which allow us to run the "text" mode. This generalizes @rvykydal [custom branch](https://github.com/rvykydal/kickstart-tests/commit/9ea056020209994dcb54ca8fa75cec8557cd1c63), which is perpetually out of date -- we shouldn't run such branches in production IMHO.

martinpitt:
I tested this locally with

    sudo containers/runner/launch --run-args '-eKSTEST_EXTRA_BOOTOPTS=text' container

This seems to have worked:
```
❱❱❱ grep -w text data/logs/kstest-container.d18rrtl_/virt-install.log 
08:00:00,616 WARNING anaconda:anaconda: stdout: Boot argument 'text' is deprecated and will be removed in the future. Please use 'inst.text' instead.
08:00:16,516 INFO anaconda:anaconda: anaconda: Display mode is set to 'interactive text mode'.
```

I also see `text` in livemedia.log.

martinpitt:
Ah, with the "deprecated" message above we shouldn't advertise "text" in README.rst. I changed it to "inst.text".

martinpitt:
Haha, that's an interesting [failure](https://travis-ci.com/github/rhinstaller/kickstart-tests/builds/203483166) -- `functions.sh` is not a test, but our current travis "which tests changed" logic trips over this.

martinpitt:
Fixed the travis test script as per the previous comment.

jkonecny12 (approved):
Looks like a nice set of improvements!

PR #416, opened 2020-11-24
Use quay.io mirror of docker.io/b4tman/squid container

martinpitt:
We constantly hit the Dockerhub pull limit from inside the Red Hat VPN
now, so docker.io registry has become useless.

### DIFF
--- a/containers/squid.sh
+++ b/containers/squid.sh
@@ -17,7 +17,7 @@ if [ "${1:-}" = start ]; then
     # This image is well-maintained (auto-built) and really small
     $CRUN run --net host --name squid --detach \
         --volume "$MYDIR"/squid-cache.conf:/etc/squid/conf.d.tail/cache.conf:ro,z \
-        --volume ks-squid-cache:/var/cache/squid docker.io/b4tman/squid
+        --volume ks-squid-cache:/var/cache/squid quay.io/rhinstaller/squid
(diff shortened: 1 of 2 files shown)


martinpitt:
I copied the image to https://quay.io/repository/rhinstaller/squid?tab=tags yesterday.

martinpitt:
FileNotFoundError: [Errno 2] No such file or directory: '/opt/kstest/kickstart-tests/containers/squid.ks.in'

Heh, need to adjust run_travis harder..

VladimirSlavik (commented):
The fix looks good, thank you!

I wonder how much organizational overhead a(nother) image to maintain adds...

martinpitt:
@VladimirSlavik : We can choose between not caring at all, and just keep that one image for as long as it works -- there is no security impact, this container is only being used inside the test environment, and has no internet access. Or we could set up a workflow which automatically copies the latest version from dockerhub to quay once a week -- this has the potential to break, but avoids falling behind for a long time. Not sure what is better really. Or maybe between Red Hat and docker we sort out the pull restriction one day, and then we can drop it again :grin:

jkonecny12:
Let's leave what we have now and switch in the future when needed.

jkonecny12 (approved):
LGTM!

PR #417, opened 2020-11-25
Pull in libvirt domain TOCTOU fix

martinpitt:
Similarly to commit 2ba1875471f, this often breaks our tests with
non-trivial parallelism. Grab the fix from GitHub as a patch and
hot-apply it in the container.

### DIFF
--- a/containers/runner/Dockerfile
+++ b/containers/runner/Dockerfile
@@ -20,10 +20,13 @@ RUN dnf -y update && \
     openssh-clients && \
     dnf -y clean all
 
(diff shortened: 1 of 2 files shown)


martinpitt:
I locally built a container with this patch on my ghtest-pitti* instances, and am now [running tests](https://github.com/martinpitt/kickstart-tests/actions/runs/382599422) on it. Previous ones still had a few "failed to run virt-install" failures, which are annoying.

jkonecny12 (approved):
Great work in catching all these TOCTOU issues. One of the hardest to find in general!

PR #418, opened 2020-11-25
Mark currently failing tests as knownfailures

martinpitt:
Let's do a clean break and mark the currently failing tests as such. Then our daily runs should ordinarily succeed, and any failure will point out an actual regression. This PR covers the tests which fail *both* on the old cobra02 infra and the containers/runner based infra.

The only remaining exception is `network-noipv4-pre`, which succeeds on cobra02 and fails *only* in the container. I am still investigating this, but will handle it as part of PR#412 or separately.

### DIFF
--- a/lvm-cache-1.sh
+++ b/lvm-cache-1.sh
@@ -17,7 +17,8 @@
 #
 # Red Hat Author(s): Chris Lumens <clumens@redhat.com>
 
-TESTTYPE="lvm storage"
+# blivet crashes with BlockDev.LVMError: https://bugzilla.redhat.com/show_bug.cgi?id=1886767
+TESTTYPE="lvm storage knownfailure"
 
 . ${KSTESTDIR}/functions.sh
 
--- a/lvm-cache-2.sh
+++ b/lvm-cache-2.sh
@@ -17,7 +17,8 @@
 #
 # Red Hat Author(s): Chris Lumens <clumens@redhat.com>
 
-TESTTYPE="lvm storage"
+# blivet crashes with BlockDev.LVMError: https://bugzilla.redhat.com/show_bug.cgi?id=1886767
+TESTTYPE="lvm storage knownfailure"
 
 . ${KSTESTDIR}/functions.sh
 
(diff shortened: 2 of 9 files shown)


martinpitt:
@poncovka : After this lands, I feel it'd be time for another "fix kickstart tests" day. This time, PRs will be self-validating thanks to running the affected tests in Travis :wink:

lvm-cache-1.sh
```diff
@@ -17,7 +17,8 @@
 #
 # Red Hat Author(s): Chris Lumens <clumens@redhat.com>
 
-TESTTYPE="lvm storage"
+# FIXME: Anaconda crashes with BlockDev.LVMError
```
poncovka, line 20: Could you please add a link to https://bugzilla.redhat.com/show_bug.cgi?id=1886767?

lvm-cache-2.sh
```diff
@@ -17,7 +17,8 @@
 #
 # Red Hat Author(s): Chris Lumens <clumens@redhat.com>
 
-TESTTYPE="lvm storage"
+# FIXME: Anaconda crashes with BlockDev.LVMError
```
poncovka, line 20: Could you please add a link to https://bugzilla.redhat.com/show_bug.cgi?id=1886767? It is the same bug as above.

poncovka (commented):
Otherwise, it looks good to me. Thank!

lvm-cache-1.sh
martinpitt: Nice that it's already supported! I adjusted the comment and commit message accordingly.

lvm-cache-2.sh
martinpitt: Dito.

martinpitt:
@poncovka : Updated, thanks! PTAL?

poncovka (approved):
Thanks!

martinpitt:
Bah, this now  had a conflict on scripts/run_travis.sh from #416 -- rebased. @poncovka , can I please ask for your re-approval?

jkonecny12 (approved):
LGTM!

poncovka (approved):

rvykydal (approved):
Looks good to me.

PR #419, opened 2020-11-25
Expose runner container's libvirt and VNC

martinpitt:
Start the user libvirt instance manually with `--listen`, so that it
listens to the TCP port. Disable TLS mode, as that requires setting up a
CA/certs; and all of this happens locally only anyway, so we don't need
authentication really.

Show a libvirt connection string to the user, so that it's easy to
control the container's libvirt on the host with `virsh -c`, or connect
to the running VMs with `virt-viewer -c` and visually follow the test
progress.

For now this only fully works with with system podman instances as they
use proper bridges.

For user podman instances we expose the port to the host, so that you
can find it in "podman ps". But this only works for controlling libvirt
itself, not for VNC.

But running as system instance is nicer anyway due to being able to use
the squid proxy, so that's not a priority for now.

### DIFF
--- a/containers/runner/launch
+++ b/containers/runner/launch
@@ -108,8 +108,10 @@ elif [ ${TEST_JOBS} -gt 4 ]; then
 fi
 
 # Run container against the local repository, to test changes easily
+# Expose the container's libvirt to the host; check "podman ps" for the port, and use e.g.:
+# virsh -c qemu+tcp://localhost:<port>/session list
 set -x
-$CRUN run -it --rm --device=/dev/kvm \
+$CRUN run -it --rm --device=/dev/kvm --publish 127.0.0.1::16509 \
     --env KSTESTS_TEST="$KSTESTS_TEST" --env TESTTYPE="${TESTTYPE:-}" --env SKIP_TESTTYPES="${SKIP_TESTTYPES:-}" \
     --env TEST_JOBS="$TEST_JOBS" ${UPDATES_IMG_ARGS:-} ${CONTAINER_RUN_ARGS:-} \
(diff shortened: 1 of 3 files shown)


jkonecny12:
This sounds great. However, why we need bridge, wouldn't it be easier just finding a port and publishing the port from the container? The IP then would always be the same.

martinpitt:
What I can do is this:
```patch
-$CRUN run -it --rm --device=/dev/kvm \
+$CRUN run -it --rm --device=/dev/kvm --publish 127.0.0.1::16509 \
```
Then `podman ps` shows a redirection to a random free port, which we can parse (through a better API than the CLI, of course):
```
a27a41773805  quay.io/rhinstaller/kstest-runner:latest          /kickstart-tests/...  2 minutes ago  Up 2 minutes ago             127.0.0.1:33515->16509/tcp  gifted_yalow
```
Then this works:
```
 Id   Name                                                      State
-------------------------------------------------------------------------
 1    kstest-container_(5bea19d3-9b52-40d3-b735-73261a7ddde5)   running
```

But virt-viewer doesn't work: "Unable to connect to the graphic server localhost:5900", as  that port only exists in the container. Need to see how I can convince libvirt to "tunnel" VNC through the libvirt port.

martinpitt:
Moving to draft for now, as this breaks user containers: They don't have an "eth0".

jkonecny12:
> Moving to draft for now, as this breaks user containers: They don't have an "eth0".

Oh, that's bad. Wish you good luck with finding solution here.

martinpitt:
@jkonecny12 : I fixed this up to not break user containers -- it just doesn't show the new message then. I added port forwarding to the host now, but as it's not very useful I didn't spend much time on making this "pretty". I'm happy to do that once we find a way to tell virt-viewer to "tunnel" VNC through libvirt instead of trying to connect directly.

jkonecny12 (approved):
Thanks a lot for this!

PR #420, opened 2020-11-26
Fix runner Dockerfile syntax for docker

martinpitt:
This file works fine with podman, but docker fails with

   When using COPY with more than one source file, the destination must be a directory and end with a /

So, do what it says.

----

See the [failed run from master branch](https://github.com/cockpit-project/cockpit/runs/1455235685?check_suite_focus=true). I ran this branch and [it succeeded](https://github.com/rhinstaller/kickstart-tests/actions/runs/384633508) and [pushed a new image](https://quay.io/repository/rhinstaller/kstest-runner?tab=tags). That's why I didn't submit this PR from my fork, but pushed the branch to origin, so that it could use the quay secrets.

### DIFF
--- a/containers/runner/Dockerfile
+++ b/containers/runner/Dockerfile
@@ -24,7 +24,7 @@ RUN dnf -y update && \
 # https://github.com/virt-manager/virt-manager/pull/179 and
 # https://github.com/virt-manager/virt-manager/pull/194 to fix parallel runs
 # This will fail once a fixed libvirt lands, remove this when that happens
-COPY ["libvirt-pool-toctou.patch", "libvirt-domain-toctou.patch", "/tmp"]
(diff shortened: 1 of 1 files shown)


martinpitt:
FTR, I disabled running Travis for pushed branches, that's just a waste. PRs are enough.

rvykydal (approved):
Looks good to me.

poncovka (approved):

jkonecny12 (approved):
LGTM! Seems like a bug on podman?

martinpitt:
@jkonecny12 : Not sure to call it a bug that podman is slightly more forgiving in its syntax.. I don't think it's actively dangerous to say "copy stuff into /some/dir" instead of "/some/dir/".

jkonecny12:
True that it's allowing the same as in Docker + something more. Which I guess is fine from the docker compatibility.

PR #421, opened 2020-11-26
Regularly run rhel8 scenario in workflow

martinpitt:
## Blockers
 - [x] PR #422
 - [x] Skip tests broken by the  [NM regression](https://bugzilla.redhat.com/show_bug.cgi?id=1903061)
 - [x] fix travis timeout, limit the number of tests that run

## [sixth run](https://github.com/rhinstaller/kickstart-tests/actions/runs/394342600)

Just one random flake, otherwise ok. (We really need PR #424!)

[Re-run without changes](https://github.com/rhinstaller/kickstart-tests/actions/runs/395577906) -- different flake, so overall every test succeeded.

## [fifth run](https://github.com/rhinstaller/kickstart-tests/actions/runs/394032530)

 - Just one ks test failure, `bridged-bond-httpks` -- forgot that in the current "skip affected tests" commit
 - Travis timed out as it tried to run too many tests (fixed in separate commit)

## [fourth run](https://github.com/rhinstaller/kickstart-tests/actions/runs/393673757)

Even more networking test failures, they just fail later on. This takes too long to fix all of them, and apparently the workaround was detrimental. Plan B: Skip these 15 tests on rhel8 and mark them in an useful way.

## [third run](https://github.com/rhinstaller/kickstart-tests/actions/runs/386618493)
- lots of networking tests fail on validation
- firewall-use-system-defaults failure, random flake

## [second run](https://github.com/rhinstaller/kickstart-tests/actions/runs/385441891)

 - lots of networking tests fail on validation

Failing Test | Failing on old cobra infra | Fails how | Fixed 
-------------|----------------|-----------------|-----------------
hello-world | Y | timed out; No such file or directory: '/usr/bin/python3' | fedora-only
https-repo | Y | no https repo defined; download.devel does not have valid cert | fedora-only
lvm-2 | Y | missing packages: python | move to python3
ntp-pools | Y | Unknown command: timesource | fedora-only| 

## [first run, for "packages" tests only](https://github.com/rhinstaller/kickstart-tests/actions/runs/385064239)

Failing Test | Failing on old cobra infra | Fails how | Fixed 
-------------|----------------|-----------------|-----------------
| container |  Y | impossible in principle with RHEL | move to fedora-only |
| unified | Y |  timeout | knownfailure
| groups-and-envs-1 | Y | timeout | move to fedora-only
| groups-and-envs-2 | Y | timeout | move to fedora-only
[...]

PR #422, opened 2020-11-27
Adjust tests for RHEL 8

martinpitt:
I went through the [first rhel8 run](https://github.com/rhinstaller/kickstart-tests/actions/runs/385441891) on the new infra, compared it with the results from cobra02. This addresses all tests which have consistently failed on both infrastructures. Mostly by skipping them on RHEL with a FIXME comment, and in one case actually fixing it. I'd really like to start the new infra runs from a clean slate, keep the tests generally succeeding, and then putting back the tests one by one with PRs that fix them.

This also adds some necessary fixes/improvements to the runner container scripts. They should not affect Fedora (tests will tell us), but they are already useful to have in master for  running RHEL 8 tests locally, until PR #421 lands.

### DIFF
--- a/harddrive-install-tree-relative.sh
+++ b/harddrive-install-tree-relative.sh
@@ -17,7 +17,9 @@
 #
 # Red Hat Author(s): Jiri Konecny <jkonecny@redhat.com>
 
-TESTTYPE=${TESTTYPE:-"packaging"}
+# FIXME: fails on RHEL 8:
+# SourceSetupError: Nothing useful found for Hard drive ISO source at partition=/dev/sdb directory=/repo/
+TESTTYPE="packaging fedora-only"
 
 . ${KSTESTDIR}/harddrive-install-tree.sh
 
--- a/unified.sh
+++ b/unified.sh
@@ -17,6 +17,7 @@
 #
 # Red Hat Author(s): Jiri Konecny <jkonecny@redhat.com>
 
-TESTTYPE="packaging rhel-only"
+# times out
+TESTTYPE="packaging rhel-only knownfailure"
 
 . ${KSTESTDIR}/functions.sh
--- a/container.sh
+++ b/container.sh
@@ -17,6 +17,6 @@
 #
 # Red Hat Author(s): Chris Lumens <clumens@redhat.com>
 
-TESTTYPE="bootloader packaging coverage"
+TESTTYPE="bootloader packaging coverage fedora-only"
 
 . ${KSTESTDIR}/functions.sh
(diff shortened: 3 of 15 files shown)


martinpitt:
Oops! No Travis right now, we [used up our monthly free capacity](https://travis-ci.com/organizations/rhinstaller/plan)  :cry: I guess I sent a few too many PRs/force pushes recently, and until yesterday we were also running tests on "push". I disabled the latter, that will at least save 50% of the capacity. (But now we need to wait for a day or so)

I'll put it on my list to find a better solution. We could use our self-hosted runners with the same restrictions that we use for the anaconda:rhel8 branch and then deploy one more instance; or maybe someone has a better idea?

martinpitt:
@poncovka : I talked to Radek about the general direction of this yesterday, and he agreed to this approach. As he's away today, do you  mind having a look at the concrete changes here? Thanks in advance!

martinpitt:
These commits were part of [this test run](https://github.com/rhinstaller/kickstart-tests/actions/runs/386618493). Most networking tests failed (unrelated, naming confusion between "ensN vs. "enpNs0"), but this PR doesn't touch any of these.

harddrive-install-tree-relative.sh
```diff
@@ -17,7 +17,9 @@
 #
 # Red Hat Author(s): Jiri Konecny <jkonecny@redhat.com>
 
-TESTTYPE=${TESTTYPE:-"packaging"}
+# FIXME: fails on RHEL 8:
+# SourceSetupError: Nothing useful found for Hard drive ISO source at partition=/dev/sdb directory=/repo/
```
poncovka, line 21: @jkonecny12 might be interested in this failure.

unified.sh
```diff
@@ -17,6 +17,7 @@
 #
 # Red Hat Author(s): Jiri Konecny <jkonecny@redhat.com>
 
-TESTTYPE="packaging rhel-only"
+# times out
```
poncovka, line 20: @jkonecny12 might be interested.

poncovka (approved):
Looks good to me. Thanks! I would go ahead and merge it. We can fix the failing tests later.

rvykydal (approved):
Looks good to me as well.

M4rtinK (approved):
Looks good to me as well - and thanks a lot! :)

harddrive-install-tree-relative.sh
jkonecny12: Making note.

unified.sh
jkonecny12: Another note.

PR #425, opened 2020-12-01
Run scenario workflow after daily-boot-iso

martinpitt:
So far the nightly run was testing the previous day's daily iso, as both
jobs ran at the same time.


jkonecny12 (approved):
LGTM!

rvykydal (approved):
Looks good to me.

martinpitt:
This is so trivial, I take the liberty to land this now, so that this evening's run will test the current image.

PR #426, opened 2020-12-03
Fix rootless podman SELinux issues

jkonecny12:
This is happening when podman is started in rootless mode. In that case /dev/kvm socket has wrong SELinux label because it just bind-mounted inside of the container. For more info see here:

https://bugzilla.redhat.com/show_bug.cgi?id=1901462

To solve this issue we need to add '--security-opt label=disabled' which will disable container SELinux separation. This will be used only if we are runnig podman in a rootless mode (use 'sudo podman' to avoid that).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@ fragments/runtime
 *.log
 *.retry
 __pycache__
+
+# these are temp directories for downloaded images and logs
+data/
--- a/containers/runner/launch
+++ b/containers/runner/launch
@@ -10,6 +10,9 @@ BASEDIR=$(dirname $(dirname $(dirname $(realpath $0))))
 TEST_JOBS=${TEST_JOBS:-$(nproc)}
 CRUN=${CRUN:-$(which podman docker 2>/dev/null | head -n1)}
 CONTAINER=quay.io/rhinstaller/kstest-runner
+# Podman in rootless mode does not have access to /dev/kvm socket https://bugzilla.redhat.com/show_bug.cgi?id=1901462
+# disable selinux container separation when podman is executed as user
+PODMAN_SELINUX_FIX=
 
 if ! test -w /dev/kvm; then
     echo "FATAL: /dev/kvm not accessible" >&2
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -3,3 +3,7 @@ fragments/runtime
 *.log
 *.retry
 __pycache__
+
+# these are temp directories for downloaded images and logs
+containers/data
```
martinpitt, line 8: This isn't meant to exist normally. It could be because you started the launch script in the containers/ directory instead of the project root, though. But then I think  just "data/" should be enough as that matches in any subdir?

containers/runner/launch
```diff
@@ -112,11 +115,16 @@ elif [ ${TEST_JOBS} -gt 4 ]; then
     VAR_TMP="-v /var/tmp"
 fi
 
+if $CRUN -v 2>&1 | grep -q "podman" && [ $(id -u) -ne 0 ]; then
```
martinpitt, line 118: A cheaper way to do this is `if [ "${CRUN%podman*}" != "$CRUN" ] ...`, avoids actually calling the program. Not a biggie, just a generally good style in shell scripts where performance matters a bit more.

martinpitt (approved):
Nice, thank you! Some nitpicks if you feel inclined, but good enough surely.

.gitignore
jkonecny12: I was thinking about it but I'm a bit concerned that it will be used on purpose. However, in that case comitter should spot that it's not part of the commit. So yes, I'll change that.

containers/runner/launch
jkonecny12: These variables changing and comparing to itself is something totally unreadable. However, yes it's good style and I should learn to use it :). Thanks for idea.

jkonecny12: Ohh this will not work because it's not podman but ``/usr/bin/podman`` and I don't want to test the complete path in case it will not be like that in some situations.

martinpitt: Yes, it's a bit unreadable if you don't know it indeed.. POSIX shell string manipulation :-/ As I said, that's only nitpicking, it's ok like this if you prefer. Otherwise, `${CRUN%podman}` is a bit cleaner indeed in this case (no `*`)

jkonecny12: My bad it's working correctly.

jkonecny12:
**UPDATED:**
* Implemented improvements pointed by @martinpitt . Thanks!

containers/runner/launch
```diff
@@ -112,11 +115,18 @@ elif [ ${TEST_JOBS} -gt 4 ]; then
     VAR_TMP="-v /var/tmp"
 fi
 
+if [ "${CRUN%podman*}" != "$CRUN" ] && [ $(id -u) -ne 0 ]; then
+    echo "Disabling SELinux container separation to enable /dev/kvm socket access."
+    PODMAN_SELINUX_FIX="--security-opt label=disable"
+fi
+
+exit 0
```
martinpitt, line 123: Debugging leftover :grin:

martinpitt (changes requested):

containers/runner/launch
jkonecny12: ohh damn! Thanks!

jkonecny12: Fixed.

martinpitt (approved):
Dekuji!

PR #428, opened 2020-12-03
Update RHEL nightly URLs

martinpitt:
The /rhel-8/ paths are a backwards compatiblity shim that is out of date
on some mirrors, so use the canonical path underneath /nightly/ instead.


martinpitt:
In particular, with the old URL, eng.brq gives a two weeks old compose, while it's right on eng.bos. With the new URL, it's right on both.

rvykydal (approved):
Looks good to me.

jkonecny12 (approved):
Yep, this is definitely nice thing for solving. Hopefully it won't break in a few months...

PR #429, opened 2020-12-07
squid caching improvements

martinpitt:


containers/squid.sh
```diff
@@ -11,9 +11,18 @@ set -eu
...
 
-if [ "${1:-}" = start ]; then
+is_running() {
+    [ -n "$($CRUN ps -q -f 'name=^squid$')" ]
+}
+
+start() {
+    if is_running; then
+        echo "Already running"
+        return 0
```
jkonecny12, line 23: Wouldn't it be better to use ``exit 0`` instead to avoid issues when the main part of this script is extended.

jkonecny12 (approved):
LGTM!

VladimirSlavik (approved):
Looks good to me. Thank you!

containers/squid.sh
martinpitt: No, I don't think so. The idea is that start() makes sure that the container is started, and that's the case if it's already running. If there is some subsequent action, then that should run no matter whether the container actually got started IMHO.

PR #430, opened 2020-12-07
Update documentation for container workflow

martinpitt:
Please let me know if we should go one step further already and stop documenting the manual way. You can still call run_kickstart_tests.sh inside the container, of course, so we can keep that part of the docs.

So let me know your preferences, please!


jkonecny12 (approved):
I agree with the current solution. Let's leave it here for some time and we can remove it later.

rvykydal (approved):
Looks good to me.

README.rst
```diff
@@ -1,22 +1,42 @@
 Kickstart Test Documentation
 ****************************
 
+Kickstart tests are one way of testing the Anaconda Installer, by running an automated installation based on a kickstart file and checking the results.
+
 :Authors:
    Chris Lumens <clumens@redhat.com>
    Martin Kolman <mkolman@redhat.com>
```
VladimirSlavik, line 8: This list is missing Martin Pitt, I think?

(Also not sure about @rvykydal's earlier contributions.)

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #432, opened 2020-12-08
Update tests for ks in stage2 on rhel

rvykydal:
In this PR I am updating only 4 tests of:
https://github.com/rhinstaller/kickstart-tests/commit/b080e2a7b57280be2f86ac260625c75b20a291bf
I'll update more in another PR.

PR #433, opened 2020-12-08
Mark pre version of vlan over bond test fedora only

rvykydal:
It is not supported on rhel.
Support on Fedora: rhbz#1879480

### DIFF
--- a/bond-vlan-pre.sh
+++ b/bond-vlan-pre.sh
@@ -17,6 +17,7 @@
 #
 # Red Hat Author(s): Radek Vykydal <rvykydal@redhat.com>
 
-TESTTYPE="network rhbz1903061"
+# Upstream support was added in https://bugzilla.redhat.com/show_bug.cgi?id=1879480
+TESTTYPE="network fedora-only"
 
 . ${KSTESTDIR}/bond-vlan-httpks.sh


bond-vlan-pre.sh
```diff
@@ -17,6 +17,6 @@
 #
 # Red Hat Author(s): Radek Vykydal <rvykydal@redhat.com>
 
-TESTTYPE="network rhbz1903061"
+TESTTYPE="network fedora-only"
```
martinpitt, line 20: Can you please add a comment pointing to https://bugzilla.redhat.com/show_bug.cgi?id=1879480 so that the "fedora-only" has some explanation attachhed? Supposedly that will trickle into rhel 9, so we can enable it then. Or perhaps it would make sense to have a `not-rhel8` tag, so that it'll automatically run on RHEL 9? (That can be a follow-up, though)

martinpitt (changes requested):
Thanks! I understand the other rhbz1903061 marked tests will need some adjustment as well, as that was not an actual bug after all?

bond-vlan-pre.sh
rvykydal: Sure, I like the idea of not-rhel8. I should do similar thing in https://github.com/rhinstaller/kickstart-tests/pull/434.

rvykydal:
> Thanks! I understand the other rhbz1903061 marked tests will need some adjustment as well, as that was not an actual bug after all?

Yes, I will check them all, the first batch is in  https://github.com/rhinstaller/kickstart-tests/pull/432

martinpitt (approved):
Thanks! Not sure if you want to do the "not-rhel-8" thing here or later, but either way is fine now. It has a proper comment now, so it's easy to figure out the history when revisiting the tests.

PR #434, opened 2020-12-08
Pre version of bridged bond is not supported on RHEL

rvykydal:
The support in upstream was added in
https://github.com/rhinstaller/anaconda/pull/2861


martinpitt (approved):
See discussion in #433, same thing applies. Thanks!

PR #437, opened 2020-12-09
Fix network gui check for virtual slaves

rvykydal:


rvykydal:
I've added the third commit which because the first one is a prerequisite to make the tests in the third commit succeed, and the first two commits would not run the relevant tests I guess (changing a included library).
Also for the first commit I am going to run regression tests for all the network tests (rhel8).

martinpitt (approved):

PR #442, opened 2020-12-10
Remove bug tag from bridged-bond-httpks tests

rvykydal:
The tests should be passing both on fedora and rhel.

### DIFF
--- a/bindtomac-bridged-bond-httpks.sh
+++ b/bindtomac-bridged-bond-httpks.sh
@@ -17,7 +17,7 @@
 #
 # Red Hat Author(s): Radek Vykydal <rvykydal@redhat.com>
 
-TESTTYPE="network rhbz1903061"
+TESTTYPE="network"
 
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
It was failing for me on the last run. But let's land that and investigate the failures (if any) in tonight's run (or just trigger one manually)

rvykydal:
> It was failing for me on the last run. But let's land that and investigate the failures (if any) in tonight's run (or just trigger one manually)

Hm, I run it in container just locally - the 2 tests only and it passed but looking at the old infra results, there are failures and successes. My gut feeling is it will be a relevant bug to be fixed.
Handling it may exercise our workflows around the new CI :)

martinpitt:
@rvykydal : Right, I did that maybe a week ago, but with several other tested changes. Great to see that it's working now!

VladimirSlavik (approved):

PR #445, opened 2020-12-11
Fix squid cleanup if container is stopped

martinpitt:
It may happen that the squid container is not running any more, but
still exists in the "exited" state. In that case, stop() will not do
anything, but removing the cache volume fails because it's still being
used by the stopped container.

Force the removal of the container in clean() to avoid that.

----

This fixes [last night's scenario failure](https://github.com/rhinstaller/kickstart-tests/actions/runs/414165831) for the two jobs that ran on ghks1.

### DIFF
--- a/containers/squid.sh
+++ b/containers/squid.sh
@@ -56,6 +56,8 @@ stop() {
 
 clean() {
     stop
+    # clean up in case it still exists, but is stopped
+    $CRUN rm -f squid 2>/dev/null || true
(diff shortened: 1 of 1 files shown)


rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
Looks good to me.

PR #446, opened 2020-12-11
Use a better workaround for lorax/podman bug

martinpitt:
Use a tmpfs volume as /var/tmp inside the container, to avoid it being
on overlayfs. This is much better than the inefficient vfs driver, and
avoids building up a parallel podman image and container store.

Thanks to Brian Lane for this!

-----

[Tested on manual run on this branch](https://github.com/rhinstaller/kickstart-tests/runs/1538955496?check_suite_focus=true). 
There are a few failures, but they already occurred [on a previous run](https://github.com/rhinstaller/kickstart-tests/runs/1532875634?check_suite_focus=true), so are not related to the broken boot.iso build (which makes *every* test fail quickly)

### DIFF
--- a/.github/workflows/daily-boot-iso-rhel8.yml
+++ b/.github/workflows/daily-boot-iso-rhel8.yml
@@ -24,8 +24,8 @@ jobs:
       - name: Build boot.iso with lorax
         run: |
           mkdir build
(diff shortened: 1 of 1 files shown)


rvykydal (approved):
Looks good to me, thank you.

Issue #447, opened 2020-12-14
bridged-bond-httpks unstable on rhel8


martinpitt:
In about 50% of runs, this test fails on the "rhel8" scenario, for example [here](https://github.com/rhinstaller/kickstart-tests/runs/1540282708?check_suite_focus=true); but it succeeded e.g. [here](https://github.com/rhinstaller/kickstart-tests/actions/runs/411753801). This reproduces fine  locally:
```
$ sudo containers/runner/launch --platform rhel8 --defaults ./scripts/defaults-rhel8.sh bridged-bond-httpks
[...]
*** Failed check: device bridge0 connected: yes
*** Failed check: bond0 has slave enp2s0 yes
*** Failed check: bond0 has slave enp3s0 yes
*** Failed check: bridge0 has ipv4 address yes

INFO: RESULT:bridged-bond-httpks:98c034d14ac0:FAILED:Validation failed with return code 1
```

[virt-install.log](https://github.com/rhinstaller/kickstart-tests/files/5686940/virt-install.log)

@rvykydal , do you have some idea about this one?

PR #450, opened 2020-12-14
Mark bridged bond httpks as r8 failure

rvykydal:


bindtomac-bridged-bond-httpks.sh
```diff
@@ -17,7 +17,7 @@
 #
 # Red Hat Author(s): Radek Vykydal <rvykydal@redhat.com>
 
-TESTTYPE="network"
+TESTTYPE="network rhel-8-failure"
```
martinpitt, line 20: Could you add a FIXME comment that refers to https://github.com/rhinstaller/kickstart-tests/issues/447 , so that it's easy to find the debug information when trying to fix this?

rvykydal: Done.

martinpitt (approved):
Cheers!

PR #451, opened 2020-12-14
Do not override TESTTYPE tags when including .sh into a test

rvykydal:


martinpitt (dismissed):
Oh, I am probably guilty of removing some of these, as I couldn't figure out what they were good for. Thanks!

rvykydal:
> Oh, I am probably guilty of removing some of these, as I couldn't figure out what they were good for. Thanks!

I don't think so, but if you are, no worries, it just shows that they the mechanism which is not great should be more obvious. But that is for another PR.

rvykydal:
Rebased for changes in https://github.com/rhinstaller/kickstart-tests/pull/450

martinpitt (approved):

PR #453, opened 2020-12-15
Clean up old infrastructure bits

martinpitt:
This brings back commit b94e277562f8 to save a considerable (~ 12 GB) amount of disk space on runners, thus improving the likelyhood that VMs can run in memory. This does not work on the old infra (I never got around to debugging why).

It also removes the ansible/linchpin scripts.

I am not sure when you want to land this, as it obviously breaks running the tests on the old infra. I mostly just want to file this as a reminder, as today is my last day. If/when this lands, you should also remove the corresponding upshift machines.


VladimirSlavik (approved):
Looks good to me. Thank you!

I'm not sure about mixing the removal and change in one PR, I see them rather as two things that depend on each other. Also technically the order might be better the other way so it always works. But never mind :-)

poncovka (approved):
Looks good to me. Thanks!

martinpitt:
@VladimirSlavik : I could file two PRs as well, but it seems pretty clear to me that (1) there needs to be a discussion about the old infra, and (2) once it gets retired, we really want both of these.  I can split it if you prefer. But my gut feeling is that @rvykydal should have the main say here.

VladimirSlavik:
Definitely!

jkonecny12 (approved):
Looks great to me! I assigned @rvykydal here to decide when we are ready for this merge or if we want to do some changes here.

martinpitt:
@rvykydal : Does this still make sense, or should I close this? Dekuji!

martinpitt:
@rvykydal : Do you still want/need this, or want to do the cleanup yourself at some point?

PR #457, opened 2021-01-06
Add one hour to nightly kickstart tests run scenario timeout

rvykydal:
The runs in the history so far take from 4 to 6 hours.  If raising the timeout
does not help much we might want to look if the runners degrade and consider
adding fresh provisioning of runner for each test.


VladimirSlavik (approved):
Looks good to me. We'll see...

PR #460, opened 2021-01-08
Master fix loop devices

jkonecny12:
Solve problems with loop devices during the Lorax builds.

Container is not able to create new loop devices so we need to pre-create them before the container with lorax is started.

Tested [here](https://github.com/Test-anaconda-org/kickstart-tests/runs/1669275525?check_suite_focus=true). It's failing but later during the workflow because 15 GiB of memory (xlarge host) is not enough for our workflow with 16 parallel test runs. Next time I should use the ocp xxl host we normally using.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #461, opened 2021-01-12
Add debugging info to container README

jkonecny12:
It's pretty hard to debug tests without this.

### DIFF
--- a/containers/runner/README.md
+++ b/containers/runner/README.md
@@ -50,6 +50,7 @@ Environment variables for the container (`--env` option):
 * KSTESTS_REPOSITORY - kickstart-tests git repository to be used
 * KSTESTS_BRANCH - kickstart-tests git branch to be used
 * BOOT_ISO - name of the installer boot iso from `data/images` to be tested (default is "boot.iso")
+* KSTEST_EXTRA_BOOTOPTS - additional boot options applied to all tests
 
 By default, the container runs the [run-kstest](./run-kstest) script. To get an
 interactive shell, append `bash` to the command line.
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Thank you, this is great!

VladimirSlavik:
On a tangent - not necessarily as part of this PR, but: There isn't any 1-2-3 howto for "_I'm a dev and I want to test my local changes with kickstart tests_". (Or I'm blind and can't see it?) The last step from debugging suggests ```--run-args="--env KSTEST_EXTRA_BOOTOPTS=inst.updates=whatever"```. But there is probably some other way, similar to how the PR test is executed. It might be useful to have that explicitly spelled out somewhere, as that is a primary reason to run a non-scheduled kickstart test.

poncovka (approved):
Looks good to me. Thanks!

containers/runner/README.md
```diff
@@ -88,3 +85,30 @@ To stop the proxy, call
...
+
+     ************************************************************************
+     You can connect to this container's libvirt with this connection string:
+    
+        qemu+tcp://${MY_IP}/session
+    
+     ************************************************************************
+     
+The last thing is to take the ``qemu+tcp://<IP>/session`` and put that to virtual 
+machine manager (``virt-manager``). Then you can see and control VM by the manager.
```
poncovka, line 111: You can run `virt-manager -c qemu+tcp://${MY_IP}/session` to add the new connection to the virtual manager.

VladimirSlavik: That's actually pretty helpful, the normal GUI way requires a choice (Hypervisor = Custom URI) that is somewhat non-obvious... I assume this also opens the window?

jkonecny12: Good point. Added.

VladimirSlavik (approved):
Thank you!

poncovka (approved):
It looks great!

PR #463, opened 2021-01-14
Grab GitHub notifications for periodic workflows

jkonecny12:
Unfortunately, GitHub will sent notifications about failures for on:schedule workflows just to the last person who did some changes to the workflow. :(

Because of that it would be great to get the notifications on me so I'm going to steal them by this PR! HAHAHAHAHAHA!


martinpitt (approved):
Officially handing over! :handshake:

poncovka (approved):
This is so silly :-D

VladimirSlavik (approved):
All your file are belong to us!

PR #468, opened 2021-01-25
Add support for modified updates images

poncovka:
Allow kickstart tests to modify provided updates images.
Check that the installer loads the provided Anaconda configuration files.

**Related to:** https://github.com/rhinstaller/kickstart-tests/pull/462

### DIFF
--- a/functions.sh
+++ b/functions.sh
@@ -190,6 +190,14 @@ start_httpd() {
     echo "${httpd_url}" > ${tmpdir}/httpd_url
 }
 
+stop_httpd() {
+    local tmpdir=$1
+
+    if [ -f ${tmpdir}/httpd-pid ]; then
+        kill $(cat ${tmpdir}/httpd-pid)
+    fi
+}
+
 start_proxy() {
     local proxy_root=$1
     local config="squid.conf"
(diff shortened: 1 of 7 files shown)


VladimirSlavik (approved):
Looks good to me, as far as I can tell... Thanks!

functions.sh
```diff
@@ -260,3 +265,41 @@ remove_iscsi_target() {
...
+upload_updates_image() {
+    local tmp_dir="${1}"
+    local updates_img="${2}"
+
+    # Copy the updates image.
+    mkdir ${tmpdir}/http
+    cp "${updates_img}" ${tmpdir}/http/updates.img
+
+    # Start the http server.
+    start_httpd ${tmpdir}/http ${tmpdir}
```
rvykydal, line 301: Maybe the test (perhaps via a function for updates image handling) should also stop the http server, as the tests starting it in .sh file do in cleanup() function.

jkonecny12 (changes requested):
Please fix the http server cleanup before merge.

functions.sh
poncovka: Fixed. However, why don't we do that by default?

poncovka:
Updated.

jkonecny12 (approved):
Looks great to me now!

PR #469, opened 2021-01-26
Add additional repository support

M4rtinK:
Make it possible to add additional repo for a kickstart test run.

At the moment we assume the repository is a regular RPM repository,
as for example created by running createrepo_c on a directory
containing a bunch of RPMs.

The main aim of this functionality is to make it possible to add
additional packages to the RPM transaction Anaconda runs during a
regular RPM based installation. Also by adding packages with a higher
version number than those in the normal package repositories it
is possible to override them with the package from the additional repo.

This can be useful to test Anaconda and/or Initial Setup scratchbuilds.

To use this feature use the -a or --additional-repo flag to the appropriate
runner script and pass either a local filesystem path to a RPM
repository or to a remote RPM repository.

If a local path is detected a localhost web server on the runner will be used to
serve the repo folder to the VM running the test. If a remote path to a
RPM repository is passed, it will be passed directly to the VM.

In both cases the additional repositories are passed via the
inst.addrepo boot option as "inst.addrepo=KSTEST_ADDITIONAL_REPOSITORY,<repo_path>".

### DIFF
--- a/containers/runner/launch
+++ b/containers/runner/launch
@@ -34,6 +34,7 @@ Options:
  -t, --testtype TYPE                  Only run TYPE tests
  -s, --skip-testtypes TYPE[,TYPE..]   Don't run tests with given types
  -u, --updates PATH|URL               Set updates.img path or URL
+ -a, --additional-repo PATH|URL       Set additional repo path or URL
  -r, --retry                          Retry failed tests once, to guard against random
                                       infrastructure failures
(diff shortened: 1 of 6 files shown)


VladimirSlavik (commented):
Looks good to me, as far as I can tell, except for one minor thing.

KKoukiou:
Let's reopen if we want to work on this again.

PR #470, opened 2021-01-26
Use correct directory path with Dockerfile

jstodola:


jkonecny12 (approved):
Thanks for this fix!

containers/runner/README.md
```diff
@@ -18,7 +18,7 @@ This will download the [automatically built](.github/workflows/container-autoupd
 
 You can also build the container yourself to test modifications to it:
 
-    podman build -t rhinstaller/kstest-runner .
+    podman build -t rhinstaller/kstest-runner containers/runner/
```
jkonecny12, line 21: Please also add --no-cache to avoid using cache and skipping the build. That would be most probably what you really want and avoid unexpected issues.

jkonecny12, line 21: To avoid problem of having to change the image path in the script we should add here ``-t quay.io/rhinstaller/kstest-runner:latest``. That should solve our problem and have it compatible with the remote image.

jkonecny12 (changes requested):
Please add `tag` to the image to prevent problem that the image is not used.

containers/runner/README.md
jstodola: Good point, fixed, thanks.

jstodola: I do not know if I like your proposal. When I list container images with `podman image ls`, I wouldn't be able to tell where the `quay.io/rhinstaller/kstest-runner` image comes from - build locally or pulled from quay.io?
I think it would be better to simply prefer the image built by the user - proposed as https://github.com/rhinstaller/kickstart-tests/pull/472

jstodola: Also, is it necessary to add the `latest` tag? The tag seems to be added automatically to the image.

jkonecny12: Yes, feel free to ignore the `latest` tag.

This is the point. There shouldn't be a difference between locally or remotely build image. The image should be the same.

jstodola:
OK, the requested changes have been implemented and this PR is ready for review.

jkonecny12 (approved):
Looks great to me now. Thanks!

PR #471, opened 2021-01-29
Additional RPM repository support

M4rtinK:
The first commit adds two library functions for working with additional RPM repositories for kickstart tests.

The second commit uses these library functions in the new Initial Setup reboot tests that need a latest Aanconda scratchbuild to work effectively.

### DIFF
--- a/functions.sh
+++ b/functions.sh
@@ -311,3 +311,61 @@ upload_updates_image() {
     # Provide the URL of the updates image.
     echo "${httpd_url}updates.img"
 }
+
+export_additional_repo() {
+    # Export additional RPM repository via localhost web server if found in the
+    # /opt/kstest/data/additional_repo well known path. We expect the folder
+    # to just contain a bunch of RPM files and will copy its contents to a tempdir
+    # and then run createrepo on it to make sure all necessary metadata is in place.
+    #
+    # Address of the localhost web server can be sourced from ${tmpdir}/addrepo_url
+    # for use in kernel_args(). Also as this starts a localhost web server do not
+    # forget to shut the server down in cleanup().
+    #
+    # This function expects one argument - path to the tempdir.
+    local tmpdir=$1
+
+    # check if additional repo exists on well known path
+
+    if [ -e /opt/kstest/data/additional_repo ]; then
+        # Copy the repo to the tmpdir so
+        # that the createrepo call odes not
+        # mess up the directory on the host.
+        # FIXME: /opt/kstest/data should either come from a env var or
+        #        runner script should copy the additional_repo folder to the
+        #        tempdir
+        cp -R /opt/kstest/data/additional_repo ${tmpdir}/kstest_additional_repo
+
+        # We expect just a bunch of RPMs, so create all the expected
+        # metadata for it.
+        createrepo_c -q ${tmpdir}/kstest_additional_repo
+
+        # log repo content for debugging purposes
+        ls ${tmpdir}/kstest_additional_repo > ${tmpdir}/additional_repo_content
+
+        # start a http server to serve the repo
+        start_httpd ${tmpdir}/kstest_additional_repo ${tmpdir}
+
+        # Store the server address in file so that it can be sourced later
+        # (eq. in kernel_args() and similar).
+        echo addrepo_url=${httpd_url} > ${tmpdir}/addrepo_url
+    fi
+}
+
+append_additional_repo_to_kernel_args() {
+    # Append boot options needed for using the additional RPM repository to
+    # the provided list of boot options and return the result.
+    #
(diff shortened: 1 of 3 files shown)


VladimirSlavik:
Umm, how is this related to https://github.com/rhinstaller/kickstart-tests/pull/469 ?

M4rtinK:
> Umm, how is this related to #469 ?

That's a good question - this is specific to supporting the Initial Setup reboot tests while #469 is a generic "I want to add a bunch of packages to the RPM transaction" support. But we are discussion maybe scrapping that for now as we have not been able to come up with some pressing needs for doing that.

Also using the functionality added by this PR one could still quite easily add packages to a specific kickstart ad-hoc, just by copy-pasting the necessary lines from the Initial Setup reboot tests & putting the packages in `data/additional_repo` folder in kickstart tests checkout root.

initial-setup-gui.sh
```diff
@@ -22,6 +22,26 @@ TESTTYPE="initial-setup"
 . ${KSTESTDIR}/functions.sh
 . ${KSTESTDIR}/validate-lib-initial-setup.sh
 
+prepare() {
+    ks=$1
+    tmpdir=$2
+    export_additional_repo $tmpdir
```
poncovka, line 28: You can probably call this in the `kernel_args` function. The `prepare` function is basically for preparing a kickstart.

initial-setup-gui.sh
```diff
@@ -22,6 +22,26 @@ TESTTYPE="initial-setup"
...
+
+kernel_args() {
+    echo $(append_additional_repo_to_kernel_args $DEFAULT_BOOTOPTS)
+}
+
+cleanup() {
+    tmpdir=$1
+
+    # cleanup the localhost webserver we might have started
+    if [ -f ${tmpdir}/httpd-pid ]; then
```
poncovka, line 40: You can call the `stop_httpd` function.

functions.sh
```diff
@@ -311,3 +311,55 @@ upload_updates_image() {
...
+    # This function expects one argument - path to the tempdir.
+    tmpdir=$1
+
+    # check if additional repo exists on well known path
+
+    if [ -e /opt/kstest/data/additional_repo ]; then
+        # Copy the repo to the tmpdir so
+        # that the createrepo call odes not
+        # mess up the directory on the host.
+        cp -R /opt/kstest/data/additional_repo ${tmpdir}/kstest_additional_repo
```
poncovka, line 334: Shouldn't the runner copy or link this directory for each test the same way as the boot.iso? I don't think that the tests should touch anything outside their temporary directory.

poncovka (approved):
I have some suggestions. Otherwise, it looks good to me. Thanks!

initial-setup-gui.sh
M4rtinK: Good point! :) That simplifies things quite a bit - thanks! :)

initial-setup-gui.sh
M4rtinK: Actually, this I left the changes in initial-setup-gui.sh by mistake, I'll drop it from there and only leave it in the Initial Setup reboot tests as we discussed earlier.

initial-setup-gui.sh
```diff
@@ -22,6 +22,26 @@ TESTTYPE="initial-setup"
...
+    export_additional_repo $tmpdir
+    echo $ks
+}
+
+kernel_args() {
+    echo $(append_additional_repo_to_kernel_args $DEFAULT_BOOTOPTS)
+}
+
+cleanup() {
+    tmpdir=$1
```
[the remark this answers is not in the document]
M4rtinK: Will do. :)

functions.sh
M4rtinK: That's what I'm not sure about as well - I've tried to find out how to properly access the data folder but have not found anything (eq. some env vars, etc.) and using the `/opt/kstest/data/`. Still definitely not ideal even just because we hardcode it in the test code effectively.

As for your suggestion, do you mean changing the runner script to detect the `additional_repo` folder in `data` and do something about it ? Eq. copy it to the `$script` path or something like that ?

I've actually got the impression we don't want to do special casing like that and tests should just fetch stuff from the `data` folder directly. Of course while making sure to use a local copy if they plan to modify the data. But I might have got it wrong. :P

M4rtinK:
PR updated:
- applied feedback from @poncovka - thanks! :)
- now dumps additional repo listing to file for easy debugging
- all seems to work fine with additional RPMs added during local testing (the IS reboot tests both run to `SUCCESS` & from the log files its apparent the Anaconda scratchbuild passed via the additional repo is being used)

poncovka (approved):
Looks good to me. Thanks!

PR #472, opened 2021-01-29
Use local container image if is exists

jstodola:
The container image created as descibed at the following page should be preferred to the official image:
https://github.com/rhinstaller/kickstart-tests/blob/master/containers/runner/README.md

### DIFF
--- a/containers/runner/README.md
+++ b/containers/runner/README.md
@@ -14,12 +14,6 @@ kickstart-tests directory in the runner container:
 
 Call `launch all` to run all tests. This can be controlled further through `--testtype` and/or `--skip-testtypes`, see `--help`.
 
-This will download the [automatically built](.github/workflows/container-autoupdate.yml) [official container image](https://quay.io/repository/rhinstaller/kstest-runner).
-
-You can also build the container yourself to test modifications to it:
-
-    podman build -t rhinstaller/kstest-runner .
-
 The `launch` script creates a `./data/` directory for passing of data between
 the container and the system (via volume).  By default it downloads the current
 Fedora Rawhide boot.iso, but to test some other image you can put it into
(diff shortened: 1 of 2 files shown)


VladimirSlavik (approved):
Thank you, good catch! I always wondered why it tries to download stuff.

(And did nothing about it, of course :-/)

VladimirSlavik (changes requested):
On a second thought, I think this means a local container will be preferred to the remote one, always? As in, even if it is a month older? That sounds like a rather major caveat: Once you build the container locally, it will forever take priority, until you somehow deal with it (update or delete). That should be documented somewhere, as it is a pretty major footgun.

Feel free to prove me wrong!

jstodola:
Exactly, a locally built container image wins - always, forever and no matter how old it is. And that's a predictable and correct behavior :-)
Really, if you build the image yourself, you probably have a good reason and you want it to be used even if there is a newer remote image available.
It's your responsibility to remove/rename the local image if you do not want to use it any more.

But I agree this behavior could be documented (a new PR), and also the `containers/runner/launch` script might print a well visible warning message when using a locally built image (update this PR). Does it sound good?

VladimirSlavik:
So, coming a few days later, I realize this is actually ok behavior-wise. The used container is printed, so that's ok. It will be in the logs.

> But I agree this behavior could be documented (a new PR)

This is not a technical requirement, but I would [personally] strongly prefer making it one PR so everything changes atomically in the repo. Multiple commits are totally fine.

I think it would be super simple to satisfy the docs side. Basically, add another paragraph after https://github.com/rhinstaller/kickstart-tests/blame/master/containers/runner/README.md#L19 to explain that the local image will be always used until you delete it with `podman remove localhost/rhinstaller/kstest-runner` or some such.

Maybe then it would be too much infromation to insert into the "main" explanation, and enough to warrant another heading `# Building the runner container locally` and moving out as a new section?

jstodola:
OK, moved the whole part with the local container image into a new section.

VladimirSlavik (approved):
Thank you - I'm happy now :)

jkonecny12 (changes requested):
This is incorrect solution. Please do not complicate scripts here.

The correct solution should be to change [README](https://github.com/rhinstaller/kickstart-tests/blob/master/containers/runner/README.md#L21) and change:

```
podman build -t rhinstaller/kstest-runner .
```

to 

```
podman build -t quay.io/rhinstaller/kstest-runner .
```

jstodola:
Please see https://github.com/rhinstaller/kickstart-tests/pull/470#discussion_r567122360 why I do not like `podman build -t quay.io/...` and proposed this solution instead.

jstodola:
Discussed with Jirka, this problem will be solved using tags for container images.
Closing this PR.

PR #478, opened 2021-02-05
Use redirector for rawhide repository

rvykydal:


VladimirSlavik (approved):
Looks good to me.

Notes: I guess there isn't any way to specify the geo like https://mirrors.fedoraproject.org/mirrorlist?repo=rawhide&arch=x86_64&country=US :( Also, pitti warned that some of the mirrors are less reliable, which was why he hardcoded a particular one.

rvykydal:
> Looks good to me.
> 
> Notes: I guess there isn't any way to specify the geo like https://mirrors.fedoraproject.org/mirrorlist?repo=rawhide&arch=x86_64&country=US :( 
Interesting idea,
http://download.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os/?&country=US works for me in the browser so we could perhaps try to use it ?
> Also, pitti warned that some of the mirrors are less reliable, which was why he hardcoded a particular one.
Yes, it was also mentioned in a comment. I returned to the redirector because the mirror we are using for the last few days stopped working yesterday.

poncovka (approved):
I like this change.

VladimirSlavik:
I actually wonder if specifying the geo is a good idea. I think it's auto-detected, so maybe it *is* better to leave it out, so that if we move the workers or something, it works well.

PR #480, opened 2021-02-08
Revert "Ignore "Call Trace:" in logs temporarily"

rvykydal:
This reverts commit 58c4b426d7ac560def350cee670850b73c72f191.

The #1893950 should be fixed now.

But we will see if failing on "Call Trace:" will introduce too much noise.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #452, opened 2020-12-15
Fix network interface naming in daily-boot-iso-rhel8 workflow

martinpitt:
For some strange reason, the path vs. slot based network interface
naming depends on the iso's CD volume ID. Move to
"RHEL-8-0-0-BaseOS-x86_64" to match the label structure of the official
iso while still avoiding the hardcoding of "8.4" (it seems it is enough
to be a digit, but which particular one does not matter).

Revert the previous commit 918215f6ce3 hack, and go back to testing both
the official and our daily image with the same defaults-rhel8.sh.

Fixes #448

----

Validated in a [manual run against this branch](https://github.com/rhinstaller/kickstart-tests/runs/1556071070?check_suite_focus=true).

Issue #448, opened 2020-12-14
network naming confusion on RHEL daily boot.iso


martinpitt:
Four network+coverage tests consistently fail on the [RHEL 8 daily scenario](https://github.com/rhinstaller/kickstart-tests/actions?query=workflow%3A%22Build+and+test+daily+RHEL+boot.iso%22):
```
onboot-activate                                         | FAILED     | error in log: packaging: base repo (URL/http://download.devel.redhat.com/nightly/rhel-8/RHEL-8/latest-RHEL-8/compose/BaseOS/x86_64/os/) not valid -- removing it
team-pre                                                | FAILED     | error in log: packaging: base repo (URL/http://download.devel.redhat.com/nightly/rhel-8/RHEL-8/latest-RHEL-8/compose/BaseOS/x86_64/os/) not valid -- removing it
network-static-2-httpks                                 | TIMED OUT  | 
bond2-httpks                                            | TIMED OUT  | 
```

When following along onboot-activate with virt-viewer, the installer shows the network device configuration screen without further progress, i.e. the kickstart does not match the available devices. 

This is because our tests ordinarily expect network device names `enpNs0`, but on the RHEL daily we have the `ens*`  pattern instead:
```
03:26:20,515 NOTICE xserver:[   304.498] Kernel command line: inst.ks=file:/onboot-activate.ks debug=1 inst.debug rd.shell=0 rd.emergency=poweroff ip=enp1s0:dhcp inst.kernel.hung_task_timeout_secs=1200 inst.stage2=hd:CDLABEL=RHEL-8-daily-0-BaseOS-x86_64
03:26:05,425 DEBUG anaconda:anaconda: network: Devices found: ['ens3', 'ens4', 'ens5']
```

and anaconda can't find the `enp*` ones:

```
03:26:06,495 WARNING org.fedoraproject.Anaconda.Modules.Network:DEBUG:anaconda.modules.network.nm_client:kickstart specification --device=enp2s0 ->  ()
03:26:06,495 WARNING org.fedoraproject.Anaconda.Modules.Network:WARNING:anaconda.modules.network.initialization:Apply kickstart: --device enp2s0 not found
03:26:06,495 WARNING org.fedoraproject.Anaconda.Modules.Network:DEBUG:anaconda.modules.network.nm_client:kickstart specification --device=enp1s0 ->  ()
[...]

PR #484, opened 2021-02-11
Set volume id for rawhide iso build

rvykydal:
Use the same volume id as the official rawhide boot.isos we are testing in
daily scenario.  Based on the id virt-install can select os specific
configuration which decides if slot or path based network naming scheme (enpX
vs ens0pY) is used.

See https://github.com/rhinstaller/kickstart-tests/issues/448
and https://github.com/rhinstaller/kickstart-tests/pull/452


VladimirSlavik (approved):
Thank you for finding the root cause! LGTM.

poncovka (approved):
Looks good to me. Thanks!

PR #490, opened 2021-02-22
Revert "Define network names only in defaults file"

rvykydal:
This reverts commit 117b5e2c1fa47a1056db5b0ccd73171aba2cdbe6
which has broken running tests on multiple hosts.

The network names need to be available also in .sh test files which was not the
case for running tests on remote hosts. The defaults file was sourced in
run_kicsktart_tests.sh and the variables were not passed by parallel to remote
hosts.

We could also pass the variables via --env option but they would have to be
passed each explicitly as "--env _" seems not to work [1]. Using single cfg file
with clear purpose seems better than having the variables listed at two places.

[1] for example by setting PARALLEL env variable before the parallel call
export PARALLEL="--env KSTEST_NETDEV1 --env KSTEST_NETDEV2 --env KSTEST_NETDEV3
                 --env KSTEST_NETDEV4 --env KSTEST_NETDEV5 --env KSTEST_NETDEV6"


VladimirSlavik (commented):
Looks good to me. Thank you!

jkonecny12 (approved):
LGTM!

PR #491, opened 2021-03-01
Do not use qxl video which seems to be broken (#1933637)

rvykydal:
Use virtio.


poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #494, opened 2021-03-16
Static unspec

rvykydal:
I am using rhel-8-failure instead of rhbz#BZNUM tag for tests disabling because the related BZ is assigned to NetworkManager and on Anaconda side the cases from the BZ are fixed upstream.


jstodola:
Thank you for the tests!

poncovka (approved):
Thanks!

rvykydal:
@poncovka I've changed the disabling of the test - added one more commit https://github.com/rhinstaller/kickstart-tests/pull/494/commits/200e5edaa196674e09ab6d72526fd34ad62ff088 (I'll squash into the 4th before merging) because it turned out the fix will be on Anaconda side and therefore we can use the rhbzNUM tagging here (it is better for tracking and re-enabling later).

poncovka (approved):

rvykydal:
Squashed after Vendy's review of added patch, as promised.

PR #495, opened 2021-03-17
Clean up the module tests

poncovka:
Remove unnecessary tests and update the installed and enabled modules.
Enable the module tests again.

Resolves: rhbz#1928049


M4rtinK (approved):
Looks good to me, thanks for the cleanup! :)
Overall it seems that the modularity effort has been loosing steam over time in Fedora, with most packages migrating away from modules back to regular packaging, mostly due to the various tricky issues that were never properly resolved. Also any development of new features related to modules basically stopped.

These two things effectively mean that a much reduced and better maintainable set of kickstart tests should be fully sufficient to check modularity support in Anaconda is still working correctly.

rvykydal (approved):
Looks good to me, thank you!
In some renamed/moved tests the #test name: comment is not the name of the file, but I am not sure we care and we can fix it later (and think over if or how it makes sense), maybe also for other tests, I am sure there are some.

poncovka:
The test names are fixed.

poncovka:
The tests are very unstable. I will check it out.

poncovka:
The tests seem to be stable now.

PR #501, opened 2021-03-29
Mark fedora-live-image-build as knownfailure (#1926632)

rvykydal:
The test runs too long (~10 mins more then 1h timeout).
See https://bugzilla.redhat.com/show_bug.cgi?id=1926632#c7

We might reenable it with higher timeout set if we add support for per-test
timout setting in the future.


VladimirSlavik (approved):
Sounds good!

PR #504, opened 2021-04-09
Per test timeout support

M4rtinK:
The first commit adds support for defining a per-test timeout override.

The second commit uses this new feature to set a 120 minute timeout for the Fedora live image build kickstart test.

A couple notes about thing's I'm not 100% sure about with this:
- what do you think about calling the bash function used for this `timeout()` ? any other suggestions ?
- what about input validation (so far we only split output from the bash functions in lists at most) ?
  - should we put the integer parsing into a try/except block and report issues or is this too niche for us to care ?
  - is the parsing to integers in the right place or would it be better to have it as a generic method in the `ShellOutput` class ?

### DIFF
--- a/scripts/launcher/run_one_test.py
+++ b/scripts/launcher/run_one_test.py
@@ -135,6 +135,9 @@ class Runner(object):
 
         kernel_args = self._shell.kernel_args()
 
+        timeout_string = self._shell.get_timeout()
+        timeout = int(timeout_string)
+
         if self._conf.updates_img_path:
             kernel_args += " inst.updates={}".format(self._conf.updates_img_path)
 
@@ -160,7 +163,7 @@ class Runner(object):
         v_conf.log_path = log_path
         v_conf.vnc = "vnc,listen=0.0.0.0"
         v_conf.boot_image = boot_args
-        v_conf.timeout = 60
+        v_conf.timeout = timeout
         v_conf.disk_paths = disk_args
         v_conf.networks = nics_args
         v_conf.runner_args = self._shell.additional_runner_args()
--- a/fedora-live-image-build.sh
+++ b/fedora-live-image-build.sh
@@ -21,3 +21,11 @@
 TESTTYPE="packaging fedora-only knownfailure"
 
 . ${KSTESTDIR}/functions.sh
+
+# This rest effectively builds a Fedora live image, which
+# among other things installs 1600+ packages and does various
+# fairly resource intensive tasks. So bump the timeout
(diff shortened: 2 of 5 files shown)


jkonecny12:
The solution is not bad  but I don't like the name `timeout`. It seems more like a callback than setter. So maybe something like `get_timeout`?

jkonecny12:
Also seems that the travis found a bug in the code.

rvykydal:
> The solution is not bad but I don't like the name `timeout`. It seems more like a callback than setter. So maybe something like `get_timeout`?

Yes I agree, moreover the function can conflict with timeout tool used in functions.sh:
https://github.com/rhinstaller/kickstart-tests/blob/b4445c885d6298b986b0f46c593900fbacf25ab4/functions.sh#L93
(eg if we add the function to the file, see my other comment)

rvykydal:
> Also seems that the travis found a bug in the code.

I think it is caused that default definition of the function is missing in functions.sh
```
get_timeout() {
    echo ""
}
```

rvykydal:
Other than the two comments I am fine with the PR as is.

poncovka:
> Other than the two comments I am fine with the PR as is.

Same here.

rvykydal:
@M4rtinK could you please look at the PR?

M4rtinK:
> @M4rtinK could you please look at the PR?

I'll take a look during next week. :)

M4rtinK:
Renamed the shell function to `get_timeout()` and added the stub to `functions.sh` - thanks for feedback! :)

jkonecny12 (approved):
Looks good to me now!

rvykydal (approved):
Looks good to me, thank you.

scripts/launcher/run_one_test.py
```diff
@@ -135,6 +135,15 @@ def _create_virtual_conf(self, log_path) -> VirtualConfiguration:
...
         kernel_args = self._shell.kernel_args()
 
+        timeout_string = self._shell.get_timeout()
+        if timeout_string:
+            # looks like the test has a custom timeout set,
+            # try to parse it to an integer
+            timeout = int(timeout_string)
+        else:
+            # no custom timeout, fall back to the default 60 minute test timeout
+            timeout = 60
```
poncovka, line 145: The default value can be provided by `functions.sh`.

M4rtinK: Nice idea, let's do that. :)

poncovka (approved):

PR #509, opened 2021-04-27
Switch bond= and ip= order in network-bootopts-bond-dhcp-httpks

jstodola:
There is an anaconda traceback in RHEL-8.3 (bug [1912898](https://bugzilla.redhat.com/show_bug.cgi?id=1912898)) when the bond= boot argument is placed before ip= on the kernel command line.
Switch the boot arguments order to cover the linked bug.

Related: rhbz#1912898

Q: do we want to change the original test or create a new test with switched arguments?

### DIFF
--- a/network-bootopts-bond-dhcp-httpks.sh
+++ b/network-bootopts-bond-dhcp-httpks.sh
@@ -23,7 +23,7 @@ TESTTYPE=${TESTTYPE:-"network"}
 
 kernel_args() {
     . ${tmpdir}/ks_url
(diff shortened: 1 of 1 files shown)


jkonecny12:
Personally I would add FIXME and changed the current one.

rvykydal (approved):
I am ok with switching the arguments.

VladimirSlavik (approved):
LGTM, but I don't think I understand enough of the context.

PR #511, opened 2021-05-05
Check for rhel 8 specifically in some of the tests

rvykydal:


poncovka (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me too.

(Why return "yes" and "no"?)

rvykydal:
> (Why return "yes" and "no"?)

:), I am not sure if my reasons are good but it is because
1) it does not matter much
2) the function replies by echo, not return (where I'd probably use 1/0)
3) the 1/0 return value in bash is not so intuitive for me
4) I thought if someone disagreed, I would finally learn the preferred way, but still - 1)

PR #518, opened 2021-05-13
Fix the proxy-cmdline test

poncovka:
The `inst.proxy` boot option is set to an empty string without this fix.


jkonecny12 (approved):
LGTM!

rvykydal (approved):
Looks good to me.

PR #520, opened 2021-05-18
Run daily test for rawhide with daily built iso

rvykydal:
We could also remove copr build of dnf from the built iso, this line:
https://github.com/rhinstaller/kickstart-tests/blob/f2f7d20d13a653d73ccf49039d8fd02a8203fbaf/.github/workflows/daily-boot-iso-rawhide.yml#L62
but it showed up to be pretty stable last few months so I'd keep it there until we see some failures caused by too unstable dnf.
Or if you like I'll remove it from 'daily-iso' and create a special scenario for dnf, eg 'daily-iso-dnf'.

### DIFF
--- a/.github/workflows/scenarios.yml
+++ b/.github/workflows/scenarios.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: [self-hosted, kstest]
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


poncovka (approved):
Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

Issue #524, opened 2021-06-04
Limit the number of VMs running in parallel based on memory


jstodola:
The number of simultaneously executed VMs is currently limited by the number of processors:
https://github.com/rhinstaller/kickstart-tests/blob/master/containers/runner/launch#L10

This logic should take into account also the available amount of memory and reduce the number of VMs running at the same time.
(For example, a system with 64 CPUs and 16 GiB of memory cannot run 64 VMs at the same time).

rvykydal:
https://github.com/rhinstaller/kickstart-tests/pull/525 has been merged.

PR #525, opened 2021-06-07
Limit number of parallel jobs based on memory

rvykydal:
https://github.com/rhinstaller/kickstart-tests/issues/524

### DIFF
--- a/containers/runner/launch
+++ b/containers/runner/launch
@@ -2,17 +2,30 @@
 # Run a set of tests (positional command line arguments) in the runner container in podman or docker.
 # Tests are taken from the current kickstart-tests checkout
 # If data/images/boot.iso exists, that is tested, otherwise it downloads the current Fedora Rawhide version.
-# This runs `nproc` parallel tests by default; this can be changed by setting $TEST_JOBS.
+# Runs --recommended-jobs number of parallel tests by default, which can be changed by setting $TEST_JOBS.
 
 set -eu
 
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

jstodola:
Thank you!

containers/runner/launch
```diff
@@ -41,12 +54,14 @@ Options:
                                        rhinstaller/kickstart-tests workflow artifacts.)
  --defaults DEFAULTS_SH_FILE          Path to file with overrides to scripts/defaults.sh
  --run-args ARGUMENTS                 Extra $CRUN options/arguments (space separated)
+ --recommended-jobs                   Get the number of jobs (--jobs option) run in parallel
+                                      by default, based on number of CPUs and RAM detected
```
jkonecny12, line 58: Could you please improve the message to be obvious what this option is doing?

Something like:
``Print number of jobs (see --jobs option) used by default and exit. These jobs are calculated from number of CPU and size of RAM.``

jkonecny12 (commented):
Looks good to me otherwise.

containers/runner/launch
rvykydal: Updated

rvykydal: @jkonecny12 - is it ok now?

jkonecny12 (approved):
Looks great to me now! Thanks a lot!

PR #526, opened 2021-06-14
Add test for ks reconfig of virtual device created by boot options

rvykydal:
Related: rhbz#1963834

The test was created as modification of
network-bootopts-bond-dhcp-httpks.  See the comment in the .ks.in for
the case it covers.

### DIFF
--- a/network-bootopts-bond-ks-override.sh
+++ b/network-bootopts-bond-ks-override.sh
@@ -1,0 +1,58 @@
+#
+# Copyright (C) 2020  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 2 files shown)


jstodola (approved):
The test fails as expected on RHEL-8.4.
Thanks for the test!

network-bootopts-bond-ks-override.sh
```diff
@@ -0,0 +1,61 @@
...
+
+inject_ks_to_initrd() {
+    echo "false"
+}
+
+cleanup() {
+    tmpdir=$1
+
+    if [ -f ${tmpdir}/httpd-pid ]; then
+        kill $(cat ${tmpdir}/httpd-pid)
```
poncovka, line 59: I think that we have a function that for this clean-up.

poncovka (approved):

network-bootopts-bond-ks-override.sh
rvykydal: Yes, updated.

poncovka (approved):

VladimirSlavik (approved):

PR #530, opened 2021-06-17
Disable network-bootoopts-bond-ks-override test temporarily on rhel8 …

rvykydal:
…(#1963834)

Disable network-bootoopts-bond-ks-override test on rhel8 until
rhbz#1963834 is fixed.


jkonecny12 (approved):
LGTM!

PR #531, opened 2021-06-17
Disable bridge-no-bootopts-net test temporarily on rawhide (#1973078)

rvykydal:
Disable bridge-no-bootopts-net test on rawhide until rhbz#1973078 is
fixed.


jkonecny12 (approved):
LGTM!

PR #535, opened 2021-06-23
Kill reboot tests after the timeout

poncovka:
Wait for reboot and shutdown of the virtual machine, but kill the test after the specified timeout.

### DIFF
--- a/reboot-initial-setup-gui.sh
+++ b/reboot-initial-setup-gui.sh
@@ -15,12 +15,14 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-TESTTYPE="reboot initial-setup gh527"
+TESTTYPE="reboot initial-setup"
 
 . ${KSTESTDIR}/functions.sh
 
 additional_runner_args() {
-    echo "--wait"
+    # Wait for reboot and shutdown of the VM,
+    # but exit after the specified timeout.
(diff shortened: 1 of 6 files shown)


reboot-initial-setup-gui.sh
```diff
@@ -15,12 +15,15 @@
...
-TESTTYPE="reboot initial-setup gh527"
+TESTTYPE="reboot initial-setup"
 
 . ${KSTESTDIR}/functions.sh
 
 additional_runner_args() {
+    # Wait for reboot and shutdown of the VM,
+    # but exit after the specified timeout.
     echo "--wait"
+    get_timeout()
```
rvykydal, line 26: I think you need to remove the brackets.
For me this would see more readable (though I am not sure how clean bash would it be)
`echo "--wait $(get_timeout)"`

poncovka: Fixed.

poncovka:
Updated and tested.

rvykydal (approved):
Looks good to me, thank you.

poncovka:
Added new checks for the initial-setup tests.
Blocked by the new initial-setup release.

rvykydal (approved):

poncovka:
Added fix for inital-setup-gui.

rvykydal:
Still looks good to me, thank you.

PR #539, opened 2021-07-01
Do not fail on gnome-kiosk which is not used on rhel 8

rvykydal:


VladimirSlavik (approved):
This does not feel specific enough, but I guess that's how things work?

poncovka (approved):
Looks good to me. Thanks!

jkonecny12 (approved):
LGTM!

PR #541, opened 2021-07-15
Disable the authconfig test for Fedora

poncovka:
The support for the authconfig tool will be removed in Fedora 35:
https://fedoraproject.org/wiki/Changes/RemoveAuthselectCompatPackage

### DIFF
--- a/authconfig.sh
+++ b/authconfig.sh
@@ -17,6 +17,6 @@
 #
 # Red Hat Author(s): Radek Vykydal <rvykydal@redhat.com>
 
-TESTTYPE="security"
+TESTTYPE="security rhel-only"
 
 . ${KSTESTDIR}/functions.sh


jstodola:
Vendy, could you keep the test for RHELs?

poncovka:
@jstodola, I have forgot. Sorry.

jkonecny12 (approved):
LGTM!

jstodola (approved):

rvykydal:
I think we are hitting the limits of our fedora-only, rhel-only and rhel-9-only tags here as used in current
https://github.com/rhinstaller/kickstart-tests/blob/master/containers/runner/scenario
With rhel-9-only tag the test would not be run for rhel8 scenario.
Given how our scenarios work - via excluding - we should probably define and use 'skip-on-fedora', 'skip-on-rhel8' and 'skip-on-rhel9' tags instead of rhel-only, fedora-only, rhel-9-only.

poncovka:
Updated. If I understand it correctly, `rhel-only` should do the job.

rvykydal (approved):
Looks good to me.
I'll try to come with a PR for using skip-X tags. It would be needed for https://github.com/rhinstaller/kickstart-tests/pull/542

rvykydal:
> Looks good to me.
> I'll try to come with a PR for using skip-X tags. It would be needed for #542

https://github.com/rhinstaller/kickstart-tests/pull/547

PR #542, opened 2021-07-15
Enable timesource ks command testing on RHEL-9

jstodola:


VladimirSlavik (approved):
Looks good to me. Thank you!

Makes me wonder if we shouldn't mark all tests by supported platforms, like here...

poncovka (approved):
Thanks!

jkonecny12 (approved):
LGTM!

jkonecny12:
It would be great to have a system to mark all rhel-9+ or something like that. However, I guess it won't be that simple.

ntp-pools.sh
```diff
@@ -18,6 +18,6 @@
 # Red Hat Author(s): Chris Lumens <clumens@redhat.com>
 
 # Fails in rhel8 with "Unknown command: timesource"
-TESTTYPE="time coverage fedora-only"
+TESTTYPE="time coverage fedora-only rhel-9-only"
```
rvykydal, line 21: fedora-only results in skipping of the test on rhel 9, see also https://github.com/rhinstaller/kickstart-tests/pull/541#issuecomment-882551874

rvykydal:
> It would be great to have a system to mark all rhel-9+ or something like that. However, I guess it won't be that simple.

That could be done by using skip-on-rhel8 as suggested in https://github.com/rhinstaller/kickstart-tests/pull/541#issuecomment-882551874

jkonecny12:
I know it is like that now but again it doesn't scale very well.

rvykydal:
The skip-on-rhel-8 tag is ready to be used (https://github.com/rhinstaller/kickstart-tests/pull/547 is merged).

rvykydal:
> @jstodola The skip-on-rhel-8 tag is ready to be used (#547 is merged).

Actually it has already been addressed in the PR.

PR #543, opened 2021-07-19
Test TUI and GUI of the initial setup in the smoke tests

poncovka:
The reboot tests can be run from pull requests in the Anaconda repository now.

### DIFF
--- a/reboot-initial-setup-gui.sh
+++ b/reboot-initial-setup-gui.sh
@@ -15,7 +15,7 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-TESTTYPE="reboot initial-setup"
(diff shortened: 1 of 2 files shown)


jkonecny12 (approved):
Looks good to me. I just wonder if one of these tests is not enough for smoke?

VladimirSlavik (approved):
Looks good to me. Thank you!

I think the price is ok. Kickstart tests are manually triggered anyway...

poncovka:
> Looks good to me. I just wonder if one of these tests is not enough for smoke?

We run different code in TUI and GUI.

rvykydal (approved):
Looks good to me. Hope rebooting will not be too flaky.

PR #544, opened 2021-07-20
Use the VirtIO disk bus

poncovka:
It should be faster than SATA and solve some of our issues.

Related: rhbz#1974337

### DIFF
--- a/autopart-encrypted-1.ks.in
+++ b/autopart-encrypted-1.ks.in
@@ -18,8 +18,8 @@ shutdown
 %end
 
 %post
-# Check if the type of /dev/sda2 is crypto_LUKS.
-crypted="/dev/sda2"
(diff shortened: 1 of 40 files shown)


VladimirSlavik (approved):
LGTM, thank you.

I wonder if there is a danger of excluding some code paths with this? I could not find anything virtio-specific in Anaconda and Blivet, so probably not in the code we intend to test.

rvykydal (approved):
Looks good to me.

poncovka:
Tested with a small subset of tests on rawhide, RHEL 8 and RHEL 9. It fixes RHEL 9 tests that are not using the LVM partitioning. Remaining tests are failing because of the `no operating system was found on this disk` error of the `virt-copy-out` when we try to copy out logs and results. I suggest to merge this PR.

jkonecny12 (approved):
LGTM!

PR #546, opened 2021-07-21
Fix the validation in the anaconda-modules test

poncovka:
The Subscription module is always activated now. It will refuse to start
if the environment doesn't provide required dependencies and services.


VladimirSlavik (approved):

rvykydal (approved):
Looks good to me, thank you.

jkonecny12 (approved):
LGTM!

PR #547, opened 2021-07-22
Use exclude tags for per-product daily scenarios

rvykydal:
We are excluding tests in scenarios (eg rhel-9) by tags, so replace the
tags specifying the scenario (eg fedora-only) by tags excluding from the
scenario (eg skip-on-rhel). This way we can also more sanely use more
fine-grained exclusion like skip-on-rhel-8.


jkonecny12:
I wonder that these tags don't need any implementation, or am I missing something?

VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal:
> I wonder that these tags don't need any implementation, or am I missing something?

They don't, they are just used by the scenario definitions according to their semantics.

VladimirSlavik:
There's one file, launcher-something, or such, that hardcodes the labels. It's easy to miss in the pile of changed tests.

containers/runner/scenario
```diff
@@ -29,7 +29,7 @@ case "$SCENARIO" in
             mkdir -p data/images
             curl -L http://download.eng.bos.redhat.com/rhel-9/development/RHEL-9-Beta/latest-RHEL-9/compose/BaseOS/x86_64/os/images/boot.iso --output data/images/boot.iso
         fi
-        $LAUNCH --skip-testtypes fedora-only,knownfailure,rhbz1960279 --platform rhel9 --defaults "$ROOTDIR/scripts/defaults-rhel9.sh" all
+        $LAUNCH --skip-testtypes skip-on-rhel,knownfailure,rhbz1960279 --platform rhel9 --defaults "$ROOTDIR/scripts/defaults-rhel9.sh" all
```
poncovka, line 32: Could we have also `skip-on-rhel-9`, please?

ntp-pools.sh
```diff
@@ -18,6 +18,6 @@
 # Red Hat Author(s): Chris Lumens <clumens@redhat.com>
 
 # Fails in rhel8 with "Unknown command: timesource"
-TESTTYPE="time coverage fedora-only"
+TESTTYPE="time coverage skip-on-rhel"
```
poncovka, line 21: This should work on RHEL 9.

poncovka (approved):
Thanks!

containers/runner/scenario
rvykydal: updated

ntp-pools.sh
rvykydal: changed to skip-on-rhel-8

poncovka (approved):

poncovka:
Btw. maybe we could get rid of the `rhel-8-failure` tag.

rvykydal:
> Btw. maybe we could get rid of the `rhel-8-failure` tag.

Yes, we should review it and maybe get rid of it, replace either with skip-on-rhel-8 or a filed rhbzX tag (but I am afraid atm it is for BZs we don't expect to get fixed in rhel8)

PR #556, opened 2021-07-30
Use https for fedoraproject URLs

siddhesh:
The https alternatives have existed for a while now so use them to be safe.


siddhesh:
OK my bad, looks like all mirrors that dl.fedoraproject.org redirects to don't always support https.  Sorry about the noise.

PR #558, opened 2021-08-04
scenarios: raise workflow timeout temporarily for debugging

rvykydal:
The daily-iso scenario seems to take ~1.7 times longer then rhel-9 since
some week ago. I am not able to reproduce the difference locally, so I'd
raise the timeout temporarily (by 1h) to allow the daily-iso to finish
so we can get the logs to see where the bottleneck can be.


jkonecny12 (approved):
Looks good to me. I guess the problem could be in the Upshift.

poncovka (approved):

PR #561, opened 2021-08-09
Do not match substrings with skip testtype feature

rvykydal:


VladimirSlavik (approved):
Looks good to me. Thank you!

scripts/run_kickstart_tests.sh
```diff
@@ -184,7 +184,7 @@ sed_args+=$(printenv | while read line; do
 function should_skip_test() {
     filepath=$1
     for testtype in $(echo "${SKIP_TESTTYPES}" | tr ',' ' '); do
-       if [[ "$(grep TESTTYPE= ${filepath})" =~ "$testtype" ]]; then
+       if [[ "$(grep ^\s*TESTTYPE= ${filepath})" =~ "$testtype" ]]; then
```
poncovka: Hi, I had some issues with adding new comments and then I forgot. Sorry. Anyway, I am not sure why would we want to match commented strings. Also, this regex will accept any line that contains `TESTTYPE=`. That doesn't seem right.

VladimirSlavik: Um, no? Start-of-string, whitespace, `TESTTYPE=`. A hash # will break that. So it does what it should. Perhaps the double negative in commit title is misleading? Do not skip commented out = ignore commented out.

VladimirSlavik: That said, now that you made me think about this more, I think it's not good enough. Here's a few lines and how they work or not:
* `TESTTYPE="rhel"` - correct before and now
* `# TESTTYPE="rhel"` - wrongly used before, correctly ignored now
* `TESTTYPE="rhel" # was also rhbz123456` - wrongly used before, wrongly used now

I'm not sure how much sense does the last option make, though, so perhaps that's good enough.

poncovka: Ok, I was confused by the description "Do not skip test types for TESTTYPE lines that are commented out". I guess the current solution is good enough. We shouldn't parse bash scripts in the first place.

jkonecny12:
Sorry for late review, but did you tried also `\W` or `[^[:word:]]` [source](https://www.boost.org/doc/libs/1_64_0/libs/regex/doc/html/boost_regex/syntax/basic_extended.html#boost_regex.syntax.basic_extended.single_character_character_class)? That would avoid adding the `,` character there.

PR #569, opened 2021-08-16
Enable Dracut visible warnings test on RHEL-9

jkonecny12:
The feature should be on RHEL-9 now so it's safe to enable it there.


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

jstodola (approved):

PR #570, opened 2021-08-16
Switch container to base on Fedora-34

jkonecny12:
Fedora-32 is now EOL. So use something still updated.

I have to fix dependencies a bit:

- The libguestfs-tools doesn't provide us tools we need.
- The createrepo works but the correct name of the package is createrepo_c so use that instead of installing it based on the provides.


poncovka (approved):

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #574, opened 2021-08-30
Add network options pre test

rvykydal:
For now the test checks only --nodefroute default option (for BZ 1990145) but it is intended to be used also for other options in the future if needed (so we save kickstart tests resources by testing multiple features in a single test).


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

PR #579, opened 2021-08-31
Add a simple anabot test

poncovka:
Introduce a simple test for interaction with GUI.

**TODO:**
- [x] Test on Fedora.
- [x] Test on RHEL 9.
- [x] Merge https://pagure.io/anabot/pull-request/21.
- [x] Merge https://github.com/rhinstaller/kickstart-tests/pull/580.
- [x] Drop the workaround with a fork.

### DIFF
--- a/anabot-1.sh
+++ b/anabot-1.sh
@@ -1,0 +1,94 @@
+#
+# Copyright (C) 2021  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+TESTTYPE="ui smoke anabot skip-on-rhel-8"
+
+. ${KSTESTDIR}/functions.sh
+
+kernel_args() {
+    local tmp_dir="${1}"
+
+    # Provide an installation source.
+    local repo_url="$( echo ${KSTEST_URL} | sed -e 's/^--url=//' )"
+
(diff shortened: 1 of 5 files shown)


anabot-1.sh
```diff
@@ -0,0 +1,87 @@
...
+    local updates_dir="${tmp_dir}/updates"
+    local updates_img="${tmp_dir}/updates.img"
+
+    local anabot_branch="main"
+    local anabot_url="https://pagure.io/anabot.git"
+    local anabot_img="${tmp_dir}/anabot.tar.gz"
+
+    # Use a fork.
+    # FIXME: Don't use the fork.
+    local anabot_url="https://pagure.io/forks/vponcova/anabot.git"
```
rvykydal, line 47: A typo, "forks" -> "fork" ?

rvykydal:
Looks good and pretty straightforward to me, thank you.

VladimirSlavik (approved):
Looks good to me. Thank you!

anabot-1.sh
poncovka: I have checked the GIT URL and it is really `forks`.

rvykydal: Ah, sorry for the noise, you are right.

anabot-1.sh
```diff
@@ -0,0 +1,87 @@
...
+
+inject_ks_to_initrd() {
+    echo "false"
+}
+
+validate() {
+    local tmp_dir="${1}"
+    copy_interesting_files_from_system "${tmp_dir}"
+    echo "SUCCESS" > "${tmp_dir}/RESULT"
+    check_result_file "${tmp_dir}"
```
jkonecny12, line 81: What is the benefit of this? Instead of creating the file and then checking it's creating, maybe we should just return 0? What are we checking here?

functions.sh
```diff
@@ -123,6 +123,7 @@ copy_interesting_files_from_system() {
     # RESULT - file from the test
     for item in /root/original-ks.cfg \
                 /root/anaconda-ks.cfg \
+                /root/anabot.log \
```
jkonecny12, line 126: Could you please fix the indentation of `\`.

jkonecny12 (commented):
I would like to ask for the below.

anabot-1.sh
VladimirSlavik: The file might be needed for something else, outside of the test itself. I think it's good to move towards "always have the file".

I agree that return 0 would work just as well for the second line.

jkonecny12: Interesting, so basically we are solving required side effect here. Could you please add a comment if that is true ideally with explanation what will happen if the side effect is not fulfilled?

poncovka: The result file now contains the anabot errors if any.

poncovka:
Updated. All temporary workarounds were removed.

VladimirSlavik (approved):
Thank you, looking forward to having more kinds of tests....

PR #580, opened 2021-08-31
Simplify the validate_RESULT function

poncovka:
Move the code to other functions.

### DIFF
--- a/functions.sh
+++ b/functions.sh
@@ -102,8 +102,15 @@ copy_file() {
 }
 
 copy_interesting_files_from_system() {
-    args="$1"
-    disksdir="$2"
+    local disksdir="$1"
+
+    # Find disks.
+    local args=$(for d in ${disksdir}/disk-*img; do echo -a ${d}; done)
+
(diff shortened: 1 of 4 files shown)


rvykydal (approved):
Looks good to me.

functions.sh
```diff
@@ -124,20 +131,18 @@ copy_interesting_files_from_system() {
 }
 
 validate_RESULT() {
-    disksdir=$1
```
VladimirSlavik, line 127: I think having it as a local named variable might have been useful, but never mind.

VladimirSlavik (approved):
Looks good to me. Thank you!

functions.sh
jkonecny12: +1 it's more readable to set a name for the variable.

jkonecny12 (approved):
Looks good to me aside the Vlada's comment.

M4rtinK (approved):
Looks good to me, nice cleanup!

PR #584, opened 2021-09-16
Add a longer wait for Xorg

VladimirSlavik:
Kickstart tests usually run en masse in VMs, which invites slow I/O.
Unfortunately, that is the leading known cause of Xorg timeouts.

### DIFF
--- a/scripts/launcher/lib/virtual_controller.py
+++ b/scripts/launcher/lib/virtual_controller.py
@@ -151,6 +151,10 @@ class VirtualInstall(object):
         if self._kernel_args:
             extra_args += " " + self._kernel_args
 
+        # work around too slow VMs failing to start Xorg
+        # TODO: Remove this again once we know how to run things normally.
+        extra_args += " inst.xtimeout=300"
+
         extra_args += " inst.stage2=hd:CDLABEL={0}".format(udev_escape(self._label))
 
         if self._boot:


jstodola:
If kickstart-tests are hitting the I/O limits, wouldn't it be better to reduce the number of VMs running in parallel instead?
I'm wondering how much longer it would take to execute kickstart-tests if the the number of parallel VMs would be reduced by 25% for example, but the performance would not be limited by I/O so much.
Also, if there is a real bug and xorg takes 1 minute to start in the installation environment, this workaround would hide this problem.

VladimirSlavik:
Yes, that is possible. However, I am concerned that what we'd do is effectively reserve some Upshift capacity with a large instance, then not use it to keep some "buffer". Would that really keep the capacity reserved?

Also, this timeout did not happen before, I think what we're seeing is Upshift getting crowded.

>  Also, if there is a real bug and xorg takes 1 minute to start in the installation environment, this workaround would hide this problem.

True. That's why this should be temporary. I changed the message and comments.

VladimirSlavik:
It seems the tests run nicely with this, not a single failure due to slow Xorg. Of course, it's hard to tell if it's the change or just luck.

https://github.com/VladimirSlavik/kickstart-tests/runs/3626842464?check_suite_focus=true

poncovka (approved):
I would merge it as a temporary fix, so we can get more logs and investigate the issue further.

PR #585, opened 2021-09-17
Set environment to access quay.io secrets

jkonecny12:
This way we will have the quay.io secrets only for the container-autoupdate workflow.


poncovka (approved):
Thanks!

VladimirSlavik (approved):
Thank you!

PR #586, opened 2021-09-17
Run only 75% tests at once, 33% longer

VladimirSlavik:
This changes how much of the host is utilized.

This is a temporary workaround that should be reverted later.

### DIFF
--- a/.github/workflows/scenarios.yml
+++ b/.github/workflows/scenarios.yml
@@ -17,9 +17,9 @@ jobs:
       fail-fast: false
 
     # these settings depend on the infrastructure; on upshift ocp-master-xxl they take about 4 hours
-    timeout-minutes: 480
+    timeout-minutes: 640
     env:
-      TEST_JOBS: 16
+      TEST_JOBS: 12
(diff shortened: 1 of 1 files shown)


poncovka (approved):
Thanks!

jkonecny12 (approved):
Looks good to me. However, it would be nice to know how much longer it would take to finish the test run with this change.

VladimirSlavik:
If everything is fluid, 25% down on parallelism means 33% up on duration. That's how I changed the numbers.

However the real "how long" depends on actually available resources, and nobody knows that in advance. I guess we might have to keep trying to run things, until they sort of work.

(I would be happy to be wrong about this.)

M4rtinK (approved):
Looks good to me as well - hope it helps. :)

rvykydal:
I am not against it but IIRC test runs over the weekend from @VladimirSlavik on a new runner seem to suggest that the current runners have degraded and reprovisioning could be better temporary mitigation. In general, the resources seem to be better during the weekend, but that would not itself explain that on the newly provisioned runner the tests run much better.

PR #590, opened 2021-09-22
Disable container test temporarily on daily-iso (#2006694)

rvykydal:
Disable container test on daily-iso until #2006694 is fixed.


VladimirSlavik (approved):
Thank you!

poncovka (approved):

PR #591, opened 2021-09-22
Use common settings for the https-repo test (#2006694)

rvykydal:
Otherwise the test would be too fragile with regard to packaging, see
bug 2006694.


VladimirSlavik (approved):
LGTM, thank you!

poncovka (approved):
Thanks!

Issue #598, opened 2021-10-08
Fix SELinux contexts test


VladimirSlavik:
With https://github.com/rhinstaller/anaconda/pull/3642, the selinux-contexts test stopped working. (Or will).

Hacking the %post script order is no longer useful. The new code runs as a task after all %post scripts. The test needs rewriting such that it operates on the resulting image.

PR #599, opened 2021-10-08
Disable selinux-contexts everywhere for now (gh#598)

VladimirSlavik:
Disable selinux-contexts temporarily on all scenarios until gh#598 is fixed.

#598


VladimirSlavik:
Rebased.

PR #603, opened 2021-10-14
Fix selinux context test

### LINKED ISSUE #598: Fix SELinux contexts test

VladimirSlavik:
Make the test independent of how anaconda does it. Instead of hacking the `%post` script order, it's now a reboot test.

Closes #598.


jkonecny12 (approved):
Looks good to me. Thanks!

PR #611, opened 2021-11-02
Ignore also /dev when checking wrong SELinux contexts

### LINKED ISSUE #607: Failing selinux-contexts on rhel-9: Would relabel /dev/vcsa5 ...
The selinux-contexts test is causing flakes and sometimes failures on rhel-9
```
2021-10-27 00:17:27,053 DEBUG: Failed to run subprocess: '['/opt/kstest/kickstart-tests/scripts/launcher/lib/launcher_interface.sh', '-i', '/opt/kstest/data/images/boot.iso', '-k', '1', '-w', '/var/tmp/kstest-selinux-contexts.2021_10_27-00_05_30.8oaqnh93', '-t', '/opt/kstest/kickstart-tests/selinux-contexts.sh', 'validate']'
stdout:
*** System has started with errors:
Would relabel /dev/vcsa5 from system_u:object_r:device_t:s0 to system_u:object_r:tty_device_t:s0
Would relabel /dev/vcsu5 from system_u:object_r:device_t:s0 to system_u:object_r:tty_device_t:s0
Would relabel /dev/vcs5 from system_u:object_r:device_t:s0 to system_u:object_r:tty_device_t:s0
Would relabel /dev/vcsa4 from system_u:object_r:device_t:s0 to system_u:object_r:tty_device_t:s0
Would relabel /dev/vcsu4 from system_u:object_r:device_t:s0 to system_u:object_r:tty_device_t:s0
Would relabel /dev/vcs4 from system_u:object_r:device_t:s0 to system_u:object_r:tty_device_t:s0
Would relabel /dev/vcsa3 from system_u:object_r:device_t:s0 to system_u:object_r:tty_device_t:s0
Would relabel /dev/vcsu3 from system_u:object_r:device_t:s0 to system_u:object_r:tty_device_t:s0
Would relabel /dev/vcs3 from system_u:object_r:device_t:s0 to system_u:object_r:tty_device_t:s0
Would relabel /dev/vcsa2 from system_u:object_r:device_t:s0 to system_u:object_r:tty_device_t:s0
Would relabel /dev/vcsu2 from system_u:object_r:device_t:s0 to system_u:object_r:tty_device_t:s0

2021-10-27 00:17:27,053 INFO: RESULT:selinux-contexts:967e24590dfd:FAILED:Validation failed with return code 1

```
[kstest.log](https://github.com/rhinstaller/kickstart-tests/files/7451932/kstest.log)
[virt-install.log](https://github.com/rhinstaller/kickstart-tests/files/7451933/virt-install.log)

VladimirSlavik:
Somehow this path disappeared during changes to the test with no explanation.

Closes #607

### DIFF
--- a/containers/runner/scenario
+++ b/containers/runner/scenario
@@ -29,7 +29,7 @@ case "$SCENARIO" in
             mkdir -p data/images
             curl -L http://download.eng.bos.redhat.com/rhel-9/development/RHEL-9-Beta/latest-RHEL-9/compose/BaseOS/x86_64/os/images/boot.iso --output data/images/boot.iso
         fi
-        $LAUNCH --skip-testtypes skip-on-rhel,skip-on-rhel-9,knownfailure,rhbz1960279,rhbz1973156,gh595,gh564,gh601,gh576,gh607 --platform rhel9 --defaults "$ROOTDIR/scripts/defaults-rhel9.sh" all
+        $LAUNCH --skip-testtypes skip-on-rhel,skip-on-rhel-9,knownfailure,rhbz1960279,rhbz1973156,gh595,gh564,gh601,gh576 --platform rhel9 --defaults "$ROOTDIR/scripts/defaults-rhel9.sh" all
         ;;
 
     # just run a single test on standard Rawhide; mostly for testing infrastructure
--- a/selinux-contexts.ks.in
+++ b/selinux-contexts.ks.in
@@ -19,7 +19,7 @@ systemctl disable selinux-autorelabel.service
 # Write the code to run after reboot. It lists files with wrong SELinux contexts.
 cat > /usr/libexec/kickstart-test.sh << 'EOF'
 
-restorecon -rvn / -e /mnt -e /proc -e /run -e /sys
+restorecon -rvn / -e /dev -e /mnt -e /proc -e /run -e /sys
 
 EOF
 
--- a/selinux-contexts.sh
+++ b/selinux-contexts.sh
@@ -17,7 +17,7 @@
 #
 # Red Hat Author(s): Vladimir Slavik <vslavik@redhat.com>
 
-TESTTYPE="security selinux gh607"
+TESTTYPE="security selinux"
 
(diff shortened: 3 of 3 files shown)


VladimirSlavik:
@jstodola, can you please review this? The previous change that removed this path is neither logical nor documented, but I think we worked on this together back then, so you might remember something I don't.

rvykydal (approved):
Looks good to me, thank you.
But I would wait for @jstodola to confirm.

VladimirSlavik:
One more thought. Perhaps the test could be made more robust by adapting the code to automatically detect filesystem types and boundaries, so that it understands what should be ignored. IIRC the SELinux scripts that run after `.autorelabel` is found during boot/startup contain a lot of logic dedicated to this exact problem. However I think it's too much work for too little gain right now.

jkonecny12:
Hmm, running `restorecon` on dev seems like something you shouldn't do. The `/dev` contains special or device files. These shouldn't be persistent which means that you should not need to fix context on them...

VladimirSlavik:
- The `-n` options makes it a dry run, almost same as with `git clean`. It only lists what would be relabeled instead of really doing it.
- The `-e something` is to exclude/ignore paths, so it will not be touched. The real path is the first one (`/`) before the bunch of `-e`

jstodola (approved):
I agree, it doesn't make much sense to check `/dev` as part of anaconda testing, the content is not created by the installer. `/dev` is excluded in our downstream tests as well.

jkonecny12:
>     * The `-n` options makes it a dry run, almost same as with `git clean`. It only lists what would be relabeled instead of really doing it.
> 
>     * The `-e something` is to exclude/ignore paths, so it will not be touched. The real path is the first one (`/`) before the bunch of `-e`

Ohh, sorry I missed the arguments you are passing there. My bad.

jkonecny12 (approved):
Looks good to me.

rvykydal (approved):
Looks good to me.

PR #616, opened 2021-11-05
Create tag for manual tests, mark the unified tests as such

VladimirSlavik:
This adds a new tag "manual" that must be ignored by all users starting kickstart tests in bulk.

Adjust all the launchers accordingly.

Soft-blocked by https://github.com/rhinstaller/anaconda/pull/3687

### DIFF
--- a/.github/workflows/daily-boot-iso-rhel8.yml
+++ b/.github/workflows/daily-boot-iso-rhel8.yml
@@ -99,7 +99,7 @@ jobs:
           cp /home/github/rhel8-daily-boot.iso data/images/boot.iso
 
       - name: Run coverage tests
-        run: sudo TEST_JOBS=16 containers/runner/launch --retry --testtype coverage --skip-testtypes skip-on-rhel,skip-on-rhel-8,knownfailure --platform rhel8 --defaults $PWD/scripts/defaults-rhel8.sh
+        run: sudo TEST_JOBS=16 containers/runner/launch --retry --testtype coverage --skip-testtypes skip-on-rhel,skip-on-rhel-8,knownfailure,manual --platform rhel8 --defaults $PWD/scripts/defaults-rhel8.sh
 
       - name: Upload test logs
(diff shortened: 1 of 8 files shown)


M4rtinK (approved):
Looks good to me, thanks!

scripts/test_manager/configurator.py
```diff
@@ -86,6 +86,8 @@ def _pre_processing_check(self, test):
         """Check if test is valid to start processing."""
         if not self._process_known_failure and test.metadata.known_failure:
             raise KnownFailureError("")
+        if test.metadata.is_manual:
```
jkonecny12, line 89: I don't think we want to block it's execution. The `manual` doesn't mean it cannot be started, only that you have to put in something custom which is not available for automatic run. In this case it's custom boot.iso which is not available on Fedora.

jkonecny12 (changes requested):
I would only remove the `manual` from the scenarios and start a question if we don't want to support also `ELN` where it should work as expected. Even if the `ELN` would have a special pick of tests it could be valuable.

jkonecny12:
I wouldn't block even the bulk execution because there could be multiple unified tests which has the same prerequisite.

scripts/test_manager/configurator.py
VladimirSlavik: Good point, sorry. I will fix that.

VladimirSlavik:
Update: Allowed running the manual tests again everywhere except the "blind" automated tests. Also rebased.

poncovka (approved):
Great idea!

jkonecny12 (approved):
Looks great to me now. Thanks!

VladimirSlavik:
@rvykydal is this ok?

rvykydal (approved):
Looks great to me, thank you.

VladimirSlavik:
Rebased to resolve the conflict.

PR #617, opened 2021-11-05
Fix module installation tests 1-4

### LINKED ISSUE #575: module-x test failing on daily-iso scenario
The module-[1234] tests started to fail on daily-iso scenario the 2021-08-26 https://github.com/rhinstaller/kickstart-tests/actions/runs/1172301273 with
```
03:37:20,524 ERR anaconda:anaconda: modules.payloads.payload.dnf.dnf_manager: Failed to disable modules!#012Problems in request:#012missing groups or modules: nodejs:15, postgresql:13
```
[virt-install.log](https://github.com/rhinstaller/kickstart-tests/files/7077504/virt-install.log)
[kstest.log](https://github.com/rhinstaller/kickstart-tests/files/7077510/kstest.log)

M4rtinK:
The unstoppable march of postgress yet again resulted in streams
that were once new and shiny being dead and forgotten and replaced
by streams that are new and shiny.

So fixup the module-1 to 4 tests with new and shiny nodejs and
postgressql stream names, fixing the tests once and for all (of November)!

Fixes #575

### DIFF
--- a/containers/runner/scenario
+++ b/containers/runner/scenario
@@ -12,7 +12,7 @@ shift
 case "$SCENARIO" in
     rawhide) $LAUNCH --skip-testtypes skip-on-fedora,knownfailure "$@" all ;;
     rawhide-text) $LAUNCH --skip-testtypes skip-on-fedora,knownfailure --run-args '-eKSTEST_EXTRA_BOOTOPTS=inst.text' "$@" all ;;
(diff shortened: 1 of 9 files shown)


rvykydal (approved):
Looks good to me.
I wonder if it would be possible to add some hint (maybe with some heuristics) about that we most probably need to update the versions of the modules so it is easier to fix by anyone when this happens again. (Digging in the test change log might be enough though, if the implementation of the hint would be too complicated or fragile).

VladimirSlavik (approved):
Looks good to me. Thank you!

Issue #618, opened 2021-11-08
daily-iso tests failing massively with new kernel: kernel NULL pointer dereference, address: 0000000000000018


rvykydal:
Reported here: https://bugzilla.redhat.com/show_bug.cgi?id=2020438
Workaround: use VGA driver: "--virtio vga"

rvykydal:
PR with temporary workaround: https://github.com/rhinstaller/kickstart-tests/pull/619

jkonecny12:
Do we want to leave this open as reminder or what is the plan here?

rvykydal:
The BZ is fixed, PR to revert the workaround: https://github.com/rhinstaller/kickstart-tests/pull/631

PR #619, opened 2021-11-08
Use vga video driver until rhbz#2020438 is fixed

rvykydal:
Tracked in rhbz#2020438 and
https://github.com/rhinstaller/kickstart-tests/issues/618


jkonecny12 (approved):
LGTM! Thanks for the fix.

M4rtinK (approved):
Looks good to me, ACK!

PR #621, opened 2021-11-09
Master fix driverdisk test

### LINKED ISSUE #605: driverdisk-disk tests timing out with no virt-install.log
The issue was imported from bugzilla to track https://bugzilla.redhat.com/show_bug.cgi?id=1973156 here.

A kickstart test is failing.

Run:  Daily run #217 (also before) (https://github.com/rhinstaller/kickstart-tests/actions/runs/944542914)
Scenario: daily-iso
Test: driverdisk-disk

The test has been failing with TIMEOUT as a flake (ie second try passed) for some time but recently it fails as "failed" (second try fails as well) more often.

Given virt-install.log is empty I guess it hangs in initramfs.
I was not able to reproduce the failure locally.

jkonecny12:
The problem was that cdrom couldn't be type of 'virtio' but 'virtio' type was added automatically. Change the solution that when the disk string starts with `path=` it will not change this string and allow complete specification.

Closes #605.

### DIFF
--- a/containers/runner/scenario
+++ b/containers/runner/scenario
@@ -12,7 +12,7 @@ shift
 case "$SCENARIO" in
     rawhide) $LAUNCH --skip-testtypes skip-on-fedora,knownfailure,manual "$@" all ;;
     rawhide-text) $LAUNCH --skip-testtypes skip-on-fedora,knownfailure,manual --run-args '-eKSTEST_EXTRA_BOOTOPTS=inst.text' "$@" all ;;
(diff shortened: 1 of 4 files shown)


jkonecny12:
Merged commits together.

rvykydal (approved):
Looks good to me, thank you!

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

M4rtinK (approved):
Looks good to me.

jkonecny12:
*UPDATED:*
* Rebased to resolve conflict.

PR #623, opened 2021-11-09
Add new driverdisk test which will test inst.dd kernel args

jkonecny12:
The driverdisk test is using kickstart command but it would be great to have both variants tested.

This will not work until https://github.com/rhinstaller/kickstart-tests/pull/621 is merged.

### DIFF
--- a/driverdisk-disk-kargs.ks.in
+++ b/driverdisk-disk-kargs.ks.in
@@ -1,0 +1,35 @@
+#version=DEVEL
+#test name: driverdisk-disk
+%ksappend repos/default.ks
+
+%ksappend common/common_no_payload.ks
+%ksappend payload/default_packages.ks
+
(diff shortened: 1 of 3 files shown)


VladimirSlavik (changes requested):
Can you use the standard includes please? I believe a lot of what you have is actually normalized. If not, please disregard this (dismiss?).

jkonecny12:
Good point. I just copied the old test which seems to have this missing.

jkonecny12:
**UPDATED:**
* Use fragments in these tests.

VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):
Looks good to me.

M4rtinK (approved):
Looks good to me as well.

PR #629, opened 2021-11-10
Re-enable basic-ftp on daily-iso

rvykydal:
rhbz#1886706 is fixed

For rhel, we might need to find a suitable ftp repo. Using Fedora one
can be very fragile, if possible at all.

### DIFF
--- a/basic-ftp.sh
+++ b/basic-ftp.sh
@@ -18,7 +18,6 @@
 #
 # Red Hat Author(s): David Shea <dshea@redhat.com>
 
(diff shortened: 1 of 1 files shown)


rvykydal:
Well, picking a reliable repo for Fedora used to be difficult either, making the test fail for a few days, but maybe I'd give it a try again.

rvykydal:
Hm, the check is failing.

jstodola (approved):
It looks OK, ftp installations are tested as part of downstream RHEL testing.

jkonecny12 (approved):
Looks good to me.

M4rtinK (approved):
Looks good to me as well. :)

VladimirSlavik (approved):
Thank you!

PR #630, opened 2021-11-10
Remove the definition of skip-testtypes from the scenario script

jstodola:
The main motivation for this change is to be able to use the defined skip-testtypes from other scripts than `containers/runner/scenario`. Specifically to make it easier to run kickstart-tests in downstream.

### DIFF
--- a/.github/workflows/daily-boot-iso-rhel8.yml
+++ b/.github/workflows/daily-boot-iso-rhel8.yml
@@ -99,7 +99,9 @@ jobs:
           cp /home/github/rhel8-daily-boot.iso data/images/boot.iso
 
       - name: Run coverage tests
-        run: sudo TEST_JOBS=16 containers/runner/launch --retry --testtype coverage --skip-testtypes skip-on-rhel,skip-on-rhel-8,knownfailure,manual --platform rhel8 --defaults $PWD/scripts/defaults-rhel8.sh
+        run: |
+          source $PWD/containers/runner/skip-testtypes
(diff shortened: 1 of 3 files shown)


rvykydal:
Thank you, moving the configuration out of the script is very nice improvement.
I prefer the Optoin 1, seems to allow even for whole line comments and commenting out tags, ie:
```
gh111
# do not skip because blah blah
# gh222
gh333
```

jkonecny12:
Great improvement. Thanks for that. I would definitely prefer to go with Option 1.

M4rtinK:
Thanks a lot from me as well! :)

And also agree that option 1 is the best & provides the best flexibility.

VladimirSlavik (approved):
Option 1 looks great, thank you!

jstodola:
OK, the change is ready for review.
One thing to consider is the placement of the `skip-testtypes` file. Currently it's `containers/runner/skip-testtypes`, but `scripts/skip-testtypes` might be better - next to `defaults*.sh` files. It's up to you.

jkonecny12 (approved):
Looks great to me. We should also change 

https://github.com/rhinstaller/anaconda/blob/master/.github/workflows/kickstart-tests.yml#L53

to use this new variant.

M4rtinK (approved):
Looks good to me as well - thanks a lot for looking into this! :)

jstodola:
Rebased

VladimirSlavik:
I think we should prioritize this, otherwise all the quick enable/disable test PRs will keep conflicting with it.

jkonecny12:
@rvykydal could you please merge this soon?

PR #625, opened 2021-11-09
Enable anabot-1 and packages-and-groups-1

poncovka:
It looks like there was a new release of the vdo package that required a newer
kernel, but only the old one was available in the development compose.

If it happens again, we might need to reconsider the development composes or
use another environment in these tests.

Resolves: gh#601, gh#564

PR #634, opened 2021-11-16
Disable anabot-1 test temporarily on daily-iso (gh#632) (gh#633)

rvykydal:
Disable anabot-1 test on daily-iso until gh#632 and gh#633 is fixed.

### DIFF
--- a/anabot-1.sh
+++ b/anabot-1.sh
@@ -15,7 +15,7 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-TESTTYPE="ui anabot skip-on-rhel-8 gh601"
+TESTTYPE="ui anabot skip-on-rhel-8 gh601 gh632 gh633"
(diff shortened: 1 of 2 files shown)


poncovka:
Hi, I am going to mark the test as a known failure. Otherwise, we would have to disable it everywhere. See: https://github.com/rhinstaller/kickstart-tests/pull/625

rvykydal:
> Hi, I am going to mark the test as a known failure. Otherwise, we would have to disable it everywhere. See: #625

Hm, okay, I wanted to get rid of knownfailure and replace it with references to issues which can be addressed where possible, and I thought in this case it is possible, especially with https://github.com/rhinstaller/kickstart-tests/pull/630 it should not hurt to have couple of references instead of catch-all knownfailure.

rvykydal:
But I can do it later for all the scenarios.

Issue #601, opened 2021-10-11
Anabot-1 test fails on rhel-9 due to package conflicts (same as packages-and-groups-1 in #564)


rvykydal:

rvykydal:
The test is failing on rhel-9 due to the same issue as tracked in https://github.com/rhinstaller/kickstart-tests/issues/564 for packages-and-groups-1. Maybe the test can be updated for another software selection ?

PR #635, opened 2021-11-16
Install only minimal package set by anabot

jstodola:
Fewer packages means less time to run the test.
Fewer packages means lower chance of hitting issues with broken rpm
dependencies (like kmod-kvdo).
Fewer packages means ~330 instead of ~1160 on RHEL.

On the other hand, there is no benefit for rawhide/Everything variant,
where the default and minimal environments install the same package set.

### DIFF
--- a/anabot-1.xml
+++ b/anabot-1.xml
@@ -16,6 +16,9 @@
       <disk name="*" action="select" />
       <done />
     </partitioning>
+    <software_selection>
+      <environment id="minimal-environment" />
+    </software_selection>
     <begin_installation />
   </hub>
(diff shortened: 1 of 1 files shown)


rvykydal (approved):
Looks good to me.

rvykydal:
Nice, this would address https://github.com/rhinstaller/kickstart-tests/issues/601, right ?

VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK (approved):
Looks good to me as well - thanks a lot! :)

PR #637, opened 2021-11-30
Make the logs passed to data volume writeable by all

rvykydal:


poncovka (approved):

rvykydal:
Actually it does not seem to work, the w+ for all is not passed during copy. Need to look at it more.

VladimirSlavik (approved):
Thank you, this was very annoying. Hopefully it helps at least with logs interpretation.

poncovka (approved):

jkonecny12 (approved):
LGTM! However, I'm missing explanation of why we need to have write access.

PR #645, opened 2022-01-07
Remove --url from the KSTEST_URL variable

poncovka:
The `KSTEST_URL` variable should contain only the URL, so it can be easily used for anything else.

### DIFF
--- a/.github/defaults.sh
+++ b/.github/defaults.sh
@@ -1,3 +1,3 @@
 # The download.fedoraproject.org automatic redirector often selects download-ib01.f.o. for GitHub's cloud, which is too unreliable; use mirror from default scripts/defaults.sh configuration
-#export KSTEST_URL='--url=http://download.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/$basearch/os/'
+#export KSTEST_URL='http://download.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/$basearch/os/'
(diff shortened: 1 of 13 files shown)


VladimirSlavik (commented):
This removes the ability to use `--mirrorlist` and `--metalink` there, so I would expect some mention of why that is not useful anymore?

poncovka:
> This removes the ability to use `--mirrorlist` and `--metalink` there, so I would expect some mention of why that is not useful anymore?

Some notes:

* We already have to drop `--url` to use the actual url from `KSTEST_URL`.
* I have prepared a bunch of new tests specifically for the `--mirrorlist` and `--metalink` options.
* Without the change, I cannot use `KSTEST_URL` to test the `repo` command, because it is using `--baseurl` instead or `--url`.
* I think we tried to use `--mirrorlist` and `--metalink` in `KSTEST_URL`, but it is too unreliable to use it as a default source.
* We can always modify the kickstart fragments to change to default source to `--mirrorlist` or `--metalink`.
* It is the only environment variable that contains a kickstart option.

@rvykydal, @jkonecny12, what do you think?

VladimirSlavik (approved):
Thanks, in that case LGTM.

jkonecny12 (approved):
I like the change. I don't think we are able to use the same link for both (url and mirrorlist) anyway but still this is more versatile solution.

rvykydal (approved):
Looks good to me, thank you.

PR #648, opened 2022-01-21
Use empty repositories in the repo-install test

poncovka:
We set up a lot of additional repositories for the test and it can fail
with the "No space left on device" error if they are too big.


rvykydal (approved):
Looks good to me, thank you.

PR #649, opened 2022-01-26
Checks whether OSCAP spoke is disabled

marta-lewandowska:
Test to check that OSCAP spoke is not shown if the OSCAP DBus
module is disabled. (RTT-3983)

### DIFF
--- a/check_for_disabled_modules.ks.in
+++ b/check_for_disabled_modules.ks.in
@@ -1,0 +1,20 @@
+# test name: check_for_disabled_addons
+
+# Use defaults
+%ksappend repos/default.ks
+%ksappend common/common_no_payload.ks
+%ksappend payload/default_packages.ks
+
+%post --nochroot
+# Check the syslog for Addons, which should have been disabled
+ 
+errors=$( grep "Anaconda.Addons*" /tmp/syslog | grep -v forbidden )
+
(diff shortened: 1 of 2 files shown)


marta-lewandowska:
Please review this new test which automates https://bugzilla.redhat.com/show_bug.cgi?id=2018954
Thank you! :)

VladimirSlavik (commented):
Thanks! I can see the value of such test, but I don't think this would work. It fails on any message that comes from addons. What makes it distinguish between Kdump and Oscap? How do we know the message actually means the modules are activated?

So, either make the grepping more specific, or rename this to be about disabled addons. Also, please link the BZ from commit too, just to make the information available.

marta-lewandowska:
Added OSCAP to the grep to make it more specific. It is a good idea!

jstodola:
I think it would be more beneficial to make the test more general, so that it can detect if any addon started, not just the oscap addon.

marta-lewandowska:
Generalized and renamed the test. Rewrote commit message to reflect new purpose.
Please review again! :)

VladimirSlavik (approved):
Thank you, looks good to me now.

rvykydal (approved):
Looks good to me. Thank you!

PR #652, opened 2022-02-02
Adapt to RHEL 9.0 boot.iso returning version 9.0 with isoinfo

rvykydal:
The fast path for Fedora now generates this (non-fatal) error message
for RHEL 9.0 boot iso:

"[: 9.0: integer expression expected"


poncovka (approved):
Thanks!

VladimirSlavik (approved):

PR #653, opened 2022-02-02
Update the repo-install test

poncovka:
Anaconda generates repo files in the same format as DNF since anaconda-36.17-1.

**Depends on:** https://github.com/rhinstaller/anaconda/pull/3811

### DIFF
--- a/repo-install.ks.in
+++ b/repo-install.ks.in
@@ -14,6 +14,19 @@ repo --name=test05 --baseurl=EMPTY_REPO_URL --includepkgs=p1,p2 --excludepkgs=p3
 
 %post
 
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
LGTM!

poncovka:
> Why not just insert "` ?`" at the right places in the script?

Do you suggest to change the regexes instead? It would work, but I would prefer to make sure that we use the new format.

VladimirSlavik:
Yes, that was the idea. But I can see why you want to special-case that.

VladimirSlavik (approved):

rvykydal (approved):
Looks good to me, thank you.

PR #655, opened 2022-02-09
Disable selinux-contests on daily-iso temporarily (rhbz#2052038)

rvykydal:
Disable selinux-contexts until rhbz#2052038 is fixed.


jkonecny12 (approved):
Looks good to me. Thanks!

Issue #660, opened 2022-02-11
repo-include, repo-exclude, repo-addrepo, repo-enable timing out on all scenarios: ERR anaconda:dnf: No match for argument: package-1


rvykydal:
It seems that repo-addrepo, repo-enable, repo-include, repo-exclude are failing for the same reason: packages not found in additional testing repo.

For repo-addrepo:
```
23:33:13,782 INFO anaconda:librepo: Downloading: http://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os//repodata/repomd.xml
23:33:13,865 INFO anaconda:librepo: Downloading: http://10.0.2.2:40000//repodata/repomd.xml
23:33:13,874 DEBUG anaconda:anaconda: modules.payloads.payload.dnf.dnf_manager: Loaded repomd.xml hashes: {'anaconda': b'\x9d?\xc8bzcJ\xe5\xee}\x8b\xfa\x81\xa1\xae\xf5\xb2\xaf\x00=\x05\xc1f\xeb>\xd1\x17\xe7\x83g\xfc\x9c', 'ADDON': b'\x0fE\xba\xbaO\xaf\xcau\xb5\xf7\x17\x9a\xd0PZS\x97n9\x16\x1f\x9a\xe6\xf6\xe5p\xc9\xd8\x96\xb3\x940'}
23:33:13,877 DEBUG anaconda:anaconda: payload.manager: Updating payload thread state: FINISHED
23:33:13,879 INFO anaconda:anaconda: threading: Thread Done: AnaPayloadThread (140417110636096)
23:33:13,902 INFO anaconda:anaconda: threading: Running Thread: AnaCheckSoftwareThread (140417192699456)
23:33:13,906 DEBUG anaconda:anaconda: ui.gui.spokes.installation_source: Clearing checks in source spoke
23:33:13,907 DEBUG anaconda:anaconda: ui.gui.spokes.installation_source: Setting up repo: ADDON
23:33:13,925 DEBUG anaconda:anaconda: modules.payloads.payload.dnf.dnf_manager: The software selection has been cleared.
23:33:13,925 DEBUG anaconda:anaconda: modules.payloads.payload.dnf.validation: Collecting selected specs.
23:33:13,928 DEBUG anaconda:anaconda: modules.payloads.payload.dnf.validation: Collecting required specs.
[...]

PR #666, opened 2022-02-15
Don't build testing rpms in a temporary directory

poncovka:
Use the local directory instead. Otherwise, the rpms will be created outside
of the testing repositories and won't be available for the installation.

Resolves https://github.com/rhinstaller/kickstart-tests/issues/660


jkonecny12 (commented):
Could you please add explanation why is that needed?

poncovka:
The commit message has changed.

jkonecny12 (approved):
Thanks a lot for the commit message now it gives sense :).

VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):
Thank you.

PR #679, opened 2022-02-22
Test the --ignorebroken option of the %packages section

poncovka:
Add a new test for `%packages --ignorebroken`.

### DIFF
--- a/packages-ignorebroken.ks.in
+++ b/packages-ignorebroken.ks.in
@@ -1,0 +1,46 @@
+#version=DEVEL
+#test name: packages-ignorebroken
+#
+# what we are testing there:
+# - that packages with dependency solving problems are ignored if --ignorebroken is used
+# - that such an installation finishes
+
+%ksappend repos/default.ks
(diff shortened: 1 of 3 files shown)


rvykydal (approved):
Thank you.

VladimirSlavik (commented):
I'm not acking (yet) because I can't tell which package should be installed, or if that matters at all.

poncovka:
> I'm not acking (yet) because I can't tell which package should be installed, or if that matters at all.

DNF will skip the optional package. That was surprising to me, because I was expecting the conflicting package to be skipped. Based on that, the test only checks that exactly one of these packages is installed at the end, so we don't have to rely on the DNF logic and interpretation.

VladimirSlavik:
Thank you! Maybe that should be written in the test? :-)

poncovka:
Updated.

VladimirSlavik (approved):
Thank you!

jkonecny12 (approved):
Looks good to me. Thanks for adding this.

PR #682, opened 2022-02-24
Allow to override KSTEST_URL and similar from environment variable

rvykydal:
This should make also integration into Permian easier.

### DIFF
--- a/scripts/defaults-rhel8.sh
+++ b/scripts/defaults-rhel8.sh
@@ -1,6 +1,6 @@
 # Default settings for testing RHEL 8. This requires being inside the Red Hat VPN.
 
 source network-device-names.cfg
-export KSTEST_URL='http://download.eng.bos.redhat.com/rhel-8/development/RHEL-8/latest-RHEL-8.6/compose/BaseOS/x86_64/os/'
-export KSTEST_MODULAR_URL='http://download.eng.bos.redhat.com/rhel-8/development/RHEL-8/latest-RHEL-8.6/compose/AppStream/x86_64/os/'
-export KSTEST_FTP_URL='ftp://ftp.tu-chemnitz.de/pub/linux/fedora/linux/development/rawhide/Everything/$basearch/os/'
+export KSTEST_URL=${KSTEST_URL:-'http://download.eng.bos.redhat.com/rhel-8/development/RHEL-8/latest-RHEL-8.6/compose/BaseOS/x86_64/os/'}
+export KSTEST_MODULAR_URL=${KSTEST_MODULAR_URL:-'http://download.eng.bos.redhat.com/rhel-8/development/RHEL-8/latest-RHEL-8.6/compose/AppStream/x86_64/os/'}
+export KSTEST_FTP_URL=${KSTEST_FTP_URL:-'ftp://ftp.tu-chemnitz.de/pub/linux/fedora/linux/development/rawhide/Everything/$basearch/os/'}
--- a/containers/runner/launch
+++ b/containers/runner/launch
@@ -148,6 +148,8 @@ fi
 set -x
 $CRUN run -it --rm --device=/dev/kvm --publish 127.0.0.1::16509 $PODMAN_SELINUX_FIX \
     --env KSTESTS_TEST="$KSTESTS_TEST" --env TESTTYPE="${TESTTYPE:-}" --env SKIP_TESTTYPES="${SKIP_TESTTYPES:-}" \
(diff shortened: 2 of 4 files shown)


scripts/defaults-rhel8.sh
```diff
@@ -1,6 +1,6 @@
 # Default settings for testing RHEL 8. This requires being inside the Red Hat VPN.
 
 source network-device-names.cfg
-export KSTEST_URL='http://download.eng.bos.redhat.com/rhel-8/development/RHEL-8/latest-RHEL-8.6/compose/BaseOS/x86_64/os/'
-export KSTEST_MODULAR_URL='http://download.eng.bos.redhat.com/rhel-8/development/RHEL-8/latest-RHEL-8.6/compose/AppStream/x86_64/os/'
-export KSTEST_FTP_URL='ftp://ftp.tu-chemnitz.de/pub/linux/fedora/linux/development/rawhide/Everything/$basearch/os/'
+export KSTEST_URL=${KSTEST_URL:-'http://download.eng.bos.redhat.com/rhel-8/development/RHEL-8/latest-RHEL-8.6/compose/BaseOS/x86_64/os/'}
```
poncovka, line 4: Couldn't Permian just generate their own `data/scripts/defaults.sh` file?

rvykydal: That was my original solution but it has its issues.
We can't generate it completely because we don't want to have the configuration (like network device naming) stored in Permian. We want to have it in kickstart-tests (at least for some time) and use it also outside / independently of Permian. So we should just override existing defaults file. Now to do that we must pick the correct one for the architecture (eg defaults-rhel8.sh) which complicates things.
I'd like to have repo(s) settings more independent of the defaults file which seems to be more platform specific. Being able to override them independently of the platform (=> defaults file) we are using.

rvykydal: Hm, you mean to load (based on --platform <PLATFORM> option) platform specific (defualts-<PLATFORM>.sh file here: https://github.com/rhinstaller/kickstart-tests/blob/bfee17ce9adb02429d78ce1abc7894fd35bb8c51/scripts/run_kickstart_tests.sh#L156 and then the override defaults file containing the urls here https://github.com/rhinstaller/kickstart-tests/blob/bfee17ce9adb02429d78ce1abc7894fd35bb8c51/scripts/run_kickstart_tests.sh#L160 via the --defaults option ( https://github.com/rhinstaller/kickstart-tests/blob/bfee17ce9adb02429d78ce1abc7894fd35bb8c51/containers/runner/launch#L128 ). That could work.

But I still think the variables from the defaults file should be overrideable from environment (https://github.com/rhinstaller/kickstart-tests/blob/master/README.rst#chapter-4-environment-variables)

VladimirSlavik (approved):
Looks good to me, thank you.

I agree this is useful even if the direction is not taken.

scripts/defaults-rhel8.sh
poncovka: We might be able to replace these variables with something more sophisticated in the future. However, it will be more difficult if we use them as API. @jkonecny12, @VladimirSlavik, what do you think?

VladimirSlavik: Personally... I think it's a valid argument, just not one that wins. Parameters or something would be more nice, but so far most of our kickstart tests really run in wrappers of wrappers of wrappers. That's hardly ideal either, but works so wins. I think this is the same :( (Feel free to disagree)

rvykydal: > We might be able to replace these variables with something more sophisticated in the future. However, it will be more difficult if we use them as API. @jkonecny12, @VladimirSlavik, what do you think?

I think that's a good point, I wanted to use the variables from inside of the permian kickstart-test module but maybe we should be careful with exposing them when we want to use some better mechanism in the future (which we need).
So I'd go with the override defaults.sh file and scratch this PR.

jkonecny12: I'm thinking that we can make one of these parameters optional. I mean when the --defaults is not specified let's guess it from the platform.

jkonecny12 (approved):
Looks good to me. Thanks!

scripts/defaults-rhel8.sh
rvykydal: > I'm thinking that we can make one of these parameters optional. I mean when the --defaults is not specified let's guess it from the platform.

We want to have the platform defaults applied (eg for network device naming) and still have possibility to override them via --defaults (eg for KSTEST_URL and similar).

Issue #667, opened 2022-02-16
rpm-ostree failing on daily-iso: program:grub2-install: error: /usr/lib/grub/i386-pc/modinfo.sh doesn't exist. Please specify --target or --directory.


rvykydal:
rpm-ostree test started to fail since Daily run Daily run #458 (https://github.com/rhinstaller/kickstart-tests/actions/runs/1835092437)

Seems there is something missing on the image we are trying to install, from the attached virt-install.log:

```
23:39:51,384 WARNING org.fedoraproject.Anaconda.Modules.Storage:INFO:program:Running in chroot '/mnt/sysroot'... grub2-install --no-floppy /dev/vda
23:39:51,511 WARNING org.fedoraproject.Anaconda.Modules.Storage:INFO:program:grub2-install: error: /usr/lib/grub/i386-pc/modinfo.sh doesn't exist. Please specify --target or --directory.
23:39:51,512 WARNING org.fedoraproject.Anaconda.Modules.Storage:DEBUG:program:Return code: 1
```
[kstest.log](https://github.com/rhinstaller/kickstart-tests/files/8079333/kstest.log)
[virt-install.log](https://github.com/rhinstaller/kickstart-tests/files/8079331/virt-install.log)

rvykydal:
PR to disable the test: https://github.com/rhinstaller/kickstart-tests/pull/668

poncovka:
The test is enabled, but we disabled the bootloader installation as a workaround.

jkonecny12:
Based on my understanding this is on purpose change on the image of IoT or in general images deployed by coreos-installer. We might be able to fix this by using Fedora SilverBlue as source.

PR #689, opened 2022-03-04
Disable the bootloader in the rpm-ostree test

poncovka:
Let's disable the bootloader installation for now, so we can keep testing
the rpm-ostree payload.

**Related:** https://github.com/rhinstaller/kickstart-tests/issues/667


VladimirSlavik (approved):
Thank you, looks good to me.

Do we need some other way to track that this is disabled - not the whole test, just some part?

rvykydal (approved):
Looks good to me.

PR #692, opened 2022-03-10
Generate json log summary

bcl:


scripts/log2json
```diff
@@ -0,0 +1,190 @@
...
+    # Ignore the virt-install cmdline
+    if "virt-install [" in line:
+        return True
+
+    return False
+
+def main(args):
+    # Read the file, a line at a time
+    # Start by ignoring lines
+    state = 0
```
jkonecny12, line 93: Could we please change this to constants or Enum? Numbers are hard to understand.

jkonecny12 (commented):
Looks good to me otherwise.

containers/runner/launch
```diff
@@ -60,12 +61,13 @@ Options:
  --recommended-jobs                   Print the number of jobs (--jobs option) used
                                       by default and exit. It is calulated from the number
                                       of CPUs and the size of RAM.
+ --scenario NAME                      Name of the running scenario, used for json logs
```
rvykydal, line 64: I think launcher should not be aware / use "scenario" concept which comes from the GitHub workflow and scenario script. I'd just use something generic like --name (or --id or maybe something better) here (and in the related variables).

rvykydal (changes requested):
Apart from the name of the option looks good to me.
Thank you!

VladimirSlavik (approved):
Thank you, looks good to me.

I understand that this is a beginning, not a finished thing?

containers/runner/launch
bcl: Sure, it's there so that the name can be added to the json summary to group them together.

bcl:
> I understand that this is a beginning, not a finished thing?

Correct. I'm still trying to sort out what the best output is so that it can be processed by whatever thing we decide to feed it into (currently looking into Splunk).

scripts/log2json
bcl: Yeah, I wasn't sure how many different states I'd have.

containers/runner/launch
bcl: I changed it to --group, so that the results can be grouped together by scenario name when imported into an analysis tool.

bcl:
I think this is ready, it will start the process of generating json summary artifacts and I'll have a PR today or tomorrow to gather those up and turn them into a weekly (or whatever time period is selected) summary. The current format is also suitable for import into log analyzing systems like Splunk.

containers/runner/launch
```diff
@@ -60,12 +61,14 @@ Options:
  --recommended-jobs                   Print the number of jobs (--jobs option) used
                                       by default and exit. It is calulated from the number
                                       of CPUs and the size of RAM.
+ --kstest-group NAME                  Name of the tests being run, used to group results in json
+                                      logs.
```
rvykydal, line 65: I don't want be to too nitpicky and I'd hate to block your work but I'd still use rather some generic option name. To me it seems that - as you say in the description of the option - it is just a name. From the launcher perspective it is not used to group results, it is used to name/tag (all) the results of this launch run. I'd name it --run-name. But I am not going to block the PR and on it as I said.

rvykydal (approved):
Looks good to me.

containers/runner/launch
bcl: Hmm, but then why not name it scenario? :) I don't really like `--run-name` since that implies that it is tied to the specific run, but it isn't since multiple runs will have the same scenario name. Naming is hard, so I'm certainly open to more suggestions. It just can't be named 'name' since that is already used for the name of the specific test for each entry.

bcl:
See #695 for an example weekly status report email/issue. The current script to generate it is a bit of a kludge, it would be cleaner if we already had a weeks worth of json reports.

bcl:
Added a script to generate a weekly summary from the artifacts and an action to run it early on Monday mornings (UTC time).
After there is a weeks worth of json reports the summary script can be updated to use them directly instead of rebuilding them from the raw log files.

scripts/weekly-summary
```diff
@@ -0,0 +1,374 @@
...
+import io
+import json
+import os
+import sys
+from subprocess import check_call
+import tempfile
+from urllib.request import Request, urlopen
+import zipfile
+
+import pycurl
```
VladimirSlavik, line 38: Is this a dependency that needs listing somewhere?

.github/workflows/weekly-summary.yml
```diff
@@ -0,0 +1,22 @@
+# Generate a weekly summary of test failures
+name: Weekly Summary
+on:
+  schedule:
+    # Run early Monday mornings, but long enough after daily test run for results
+    - cron: 0 4 * * 1
```
VladimirSlavik, line 6: I'd suggest `0 8 * * 1`, to ensure this runs late enough to catch also the "everything timed out" case.

The timeout for each job is 480 minutes = 8 hours, there are 3 workers to take 3 jobs from the matrix, things start at 23:00, so that would give a buffer of 1 hour for starting one of the runners later because somebody was in late or some such. Given when we usually start the work day, having results at 4 or 8 AM should be the same... @rvykydal?

VladimirSlavik (commented):
Still LGTM, found two things to consider maybe.

containers/runner/launch
rvykydal: I agree that if we go concrete here, then the --scenario is better :)

scripts/weekly-summary
```diff
@@ -0,0 +1,374 @@
...
+
+URL = "https://api.github.com/repos/rhinstaller/kickstart-tests/actions/artifacts?per_page=100"
+
+# TODO - only works after we have a week of json summary runs
+#ARTIFACTS = ["summary-daily-iso", "summary-rhel9", "summary-rhel8"]
+
+# For now use the full log artifacts
+ARTIFACTS = ["logs-daily-iso", "logs-rhel9", "logs-rhel8"]
```
rvykydal, line 46: I think it would be good to be able to pass this as an argument to the script so that the eventual updates can be done in a workflow. Perhaps --artifacts=logs-daily-iso,logs-rhel9,logs-rhel8

scripts/weekly-summary
```diff
@@ -0,0 +1,374 @@
...
+        updated_at = datetime.fromisoformat(a["updated_at"][:-1])
+        datename = a["name"]+updated_at.strftime("-%Y-%m-%d")
+        filename = datename + ".zip"
+        zipfiles.append((a["name"], datename, filename))
+        if os.path.exists(filename):
+            print(f"{filename} skipped, already downloaded")
+            continue
+
+        print(f"Fetching {filename}")
+        run_curl(token, a["archive_download_url"], filename)
```
rvykydal, line 148: Maybe the return value could be checked here but reacting to it would complicate the things quite a lot at this stage so it can be eventually done when / if we really hit some failures.

scripts/weekly-summary
```diff
@@ -0,0 +1,374 @@
...
+            print(f"    {g}:", file=buf)
+            for n in sorted(days[day][g]):
+                print(f"        {n}", file=buf)
+        print("\n", file=buf)
+
+    # Gather up detailed failure results
+    for g in sorted(all_data.keys()):
+        print("=" * 40, file=buf)
+        total, good, bad = pass_fail(all_data[g])
+        print(f"{g} summary: Ran {total} tests. {good} passed, {bad} failed.", file=buf)
```
rvykydal, line 289: I think it would be useful to work with concepts of failure / flake here. I'll go into detail in a standalone comment.

VladimirSlavik: Oh my, yes please. Maybe not in this very PR, but eventually we want that pretty badly I think.

That's the main problem of current reporting - retries on failure count two results, which tells me nothing. I want to know "succeeded on retry", not have to parse the logs with my own brain to arrive at that conclusion :-) Also, failures always retry, so for real failures that "correctly re-fail", the final number is artificially inflated (IIRC).

It's one of the reasons why I consider the aggregate numbers after a kickstart tests run meaningless, unless it's a clean success, and instead look at the logs.

rvykydal (changes requested):
Looks great to me, for this PR I'd just pass the list of artifact names to weekly_report via argument.

I have some ideas regarding the summary (`logs-daily-iso: Ran 1534 tests. 1501 passed, 33 failed.`) :
1) Maybe I'd work with a concept of flake. We run the tests with --retry option which means in case of a test failure try to run the test again once more. We consider test that failed and then succeeded as a success (with flake). And I think this could be also visible in the summary. This is what I am using when creating current summary in permian (parsing the output / tests.log of the test run) (https://github.com/rvykydal/kickstart-tests/runs/5497126774?check_suite_focus=true, https://github.com/rhinstaller/permian/blob/339bcfda8269e25e40ad12db39ad2a812730a214/libpermian/plugins/kickstart_test/__init__.py#L116):
```
SUCCESS: 186 FAILED: 2 N/A: 33 (runs: 192 flakes: 2 timed out: 1)
```
`SUCCESS` - number of tests that passed (including flakes, ie fail,pass)
`FAILED` - number of tests that failed twice, ie excluding flake failures
`N/A` - tests for which we don't have a result yet
`runs` - number of all test runs (Ran: in your summary)
`flakes` - number of flakes (fail,pass of a test)
`timed out` - this info is also useful indicator sometimes

2) I think it could be useful to have such summary also for each day/scenario, eg
```
03/16/2022
    logs-daily-iso (SUMMARY):
```
It can became very useful for quick picture when there are mass failures, workflow timeouts, infra issues etc.


But we don't need to solve these in this patch set, we can discuss it in another separate PRs / issues. Maybe the summary script could have --retry option to count with flakes.

scripts/weekly-summary
rvykydal: > It's one of the reasons why I consider the aggregate numbers after a kickstart tests run meaningless, unless it's a clean success, and instead look at the logs.

It should become better when we run it via permian, see the summary added after the aggregate numbers (and also each test finish): https://github.com/rvykydal/kickstart-tests/runs/5497126774?check_suite_focus=true

scripts/weekly-summary
bcl: Oh good point.

scripts/weekly-summary
bcl: right, run_curl actually logs the errors, and I want to continue with the other logs so there is no need to check, even though it does return the status.

bcl:
Also, Simon mentioned that we should be able to trigger the actions by selecting the branch in the action page. I don't seem to have permission to do this. I'd like to see the weekly summary action run once if someone can get it to go before merging this.

rvykydal (approved):
Looks good to me.

rvykydal:
> Also, Simon mentioned that we should be able to trigger the actions by selecting the branch in the action page. I don't seem to have permission to do this. I'd like to see the weekly summary action run once if someone can get it to go before merging this.

AFAIK there is an issue that prevents triggering of workflows (having workflow_dispatch) in branches manually in UI: the workflow has to be present on deafult branch so that it listed in Actioins in GitHub UI.

I was trying to overcome it trying to run the workflow via API but it was kind of clumsy (not quite sure now if it really worked).
So what we do to test workflows is run them on our fork on default (master) branch (or have a dummy workflow of the name on default and run the workflow in branch but it is not big improvement). Or it should possible to change default branch temporarily but I haven't tried this way.

.github/workflows/weekly-summary.yml
```diff
@@ -0,0 +1,23 @@
...
+  weekly_summary:
+    name: Create weekly summary of test runs
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+      - name: Install pycurl
+        run: apt-get install -y python3-pycurl
```
rvykydal, line 20: Needs to be run as root (I hit it when tried to run the workflow locally).
`run: sudo apt-get install -y python3-pycurl`

scripts/log2json
```diff
@@ -0,0 +1,195 @@
...
+            m = re.match(r"(\d+-\d+-\d+\s\d+:\d+:\d+,\d+)", line)
+            timestamp = None
+            if m is not None:
+                try:
+                    timestamp = mktime(strptime(m.group(), "%Y-%m-%d %H:%M:%S,%f"))
+                except ValueError:
+                    pass
+
+            if m is not None:
+                line = line.removeprefix(m.group()).strip()
```
rvykydal, line 113: repmoveprefix would be a problem running on runner having python 3.8 (https://github.com/rvykydal/kickstart-tests/runs/5689533247?check_suite_focus=true)

rvykydal:
I have one more idea. Again, for another potential follow-up PR. At this stage, for human investigating of the failures, it would be very useful if the workflow produced artifacts with logs of failed tests with the same directory structure as the results (say from the results in https://github.com/rhinstaller/kickstart-tests/issues/695 ):
```
03/15/2022
    logs-rhel9:
        reboot-initial-setup-tui
        repo-install
        services


03/16/2022
    logs-daily-iso:
        bindtomac-network-device-mac
        container
        harddrive-install-tree
        hostname
        network-static-2-httpks
        network-static-2-httpks
        packages-instlangs-2
        selinux-enforcing
    logs-rhel9:
        network-addr-gen-mode
        network-prefixdevname
        repo-enable


03/17/2022
    logs-daily-iso:
        keyboard-convert-x-override-bootopt
        network-bootopts-static-unspec-bootif
    logs-rhel9:
        firewall
        network-static-to-dhcp-pre-single
        vlan-pre


03/18/2022
    logs-daily-iso:
        lvm-thinp-1
        ntp-nontp-without-chrony-gui
        onboot-activate-httpks
        rpm-ostree
    logs-rhel8:
        bindtomac-bridge-2devs-pre
    logs-rhel9:
        bindtomac-network-static-2-httpks
        bindtomac-network-static-2-httpks


03/19/2022
    logs-daily-iso:
        bindtomac-bond2-pre
        bootloader-3
        network-device-default-pre-hostname
        network-static-2-pre
    logs-rhel8:
        bindtomac-team-httpks
    logs-rhel9:
        keyboard-convert-x-override-bootopt


03/20/2022
    logs-daily-iso:
        bindtomac-bridge-2devs-pre
        default-fstype
        module-4
        network-static-httpks
    logs-rhel9:
        network-device-mac
        network-missing-ifcfg-httpks


03/21/2022
    logs-daily-iso:
        clearpart-2
        default-systemd-target-startxonboot
        module-2
        network-device-mac-pre
    logs-rhel9:
        lvm-thinp-1
```

bcl:
I tried to get the action to run in my branch and totally failed to figure it out. I think the 2 issues above are now fixed.

rvykydal:
> I tried to get the action to run in my branch and totally failed to figure it out. I think the 2 issues above are now fixed.

The last version seems to be working great, I run it as bcl2 branch in my fork: https://github.com/rvykydal/kickstart-tests/runs/5721098734?check_suite_focus=true

rvykydal (approved):
Looks good to me.

rvykydal:
> I tried to get the action to run in my branch and totally failed to figure it out. I think the 2 issues above are now fixed.

Maybe my hints were confusing. It should not be very complicated.
- on your fork of anaconda repo push the workflow to master branch (so it is visible as action in the GitHub UI)
- in your developer settings generate a GitHub personal access token with public_repo privs (so you can dowlnoad the logs from rhinstaller repo)
- add the token to your fork's anaconda repo as a secret
- update the workflow to use the secret
- run the workflow / action from the UI on your branch

bcl:
Great, let's get this merged and then I'll work on tweaking the report.

It looks like the actions in my fork started trying to run over the weekend :) So whatever I did right seems to have had a delayed response.

rvykydal:
> Great, let's get this merged and then I'll work on tweaking the report.

I'd give opportunity to @jkonecny12 to look at it and then merge.

scripts/log2json
```diff
@@ -0,0 +1,194 @@
...
+
+    return False
+
+def main(args):
+    # Read the file, a line at a time
+    # Start by ignoring lines
+    capture_summary = False
+    tests = []
+    entry = {}
+    stats = {"total": 0, "pass": 0, "fail": 0}
```
jkonecny12, line 91: What is the purpose of this variable? I don't see it used anywhere.

scripts/weekly-summary
```diff
@@ -0,0 +1,374 @@
...
+        for k, v in options.items():
+            c.setopt(k, v)
+
+        try:
+            c.perform()
+            ok = (True, None)
+        except Exception as e:
+            print(f"ERROR: {e}", file=sys.stderr)
+            sys.stderr.flush()
+            ok = (False, e)
```
jkonecny12, line 129: I think an exception is more appropriate here but I don't mind.

scripts/weekly-summary
```diff
@@ -0,0 +1,374 @@
...
+
+
+def pass_fail(tests):
+    """
+    pass_fail returns stats on the tests
+
+    It returns a tuple of total tests run, number passed, number failed
+    """
+    fail = 0
+    not_fail = 0
```
jkonecny12, line 197: I would rather name this `passed` instead of `not_fail`.

Also this is not necessary. When you are returning total tests runs that it's easy to compute that.

scripts/weekly-summary
```diff
@@ -0,0 +1,374 @@
...
+            if day not in days:
+                days[day] = {g: [t["name"]]}
+            else:
+                if g not in days[day]:
+                    days[day][g] = [t["name"]]
+                else:
+                    days[day][g].append(t["name"])
+
+    for day in sorted(days.keys()):
+        if day == "19691231":
```
jkonecny12, line 277: What is this constant?

scripts/weekly-summary
```diff
@@ -0,0 +1,374 @@
...
+            help="Rebuild logs with timestamps")
+    parser.add_argument("--output",
+            help="Path and filename to write summary report to")
+    parser.add_argument("--debug", default=False, action="store_true")
+    parser.add_argument("--verbose", default=False, action="store_true")
+    args = parser.parse_args()
+
+    if "GITHUB_TOKEN" not in os.environ:
+        print("Set GITHUB_TOKEN environmental variable to github token with access to the artifact api.")
+        sys.exit(1)
```
jkonecny12, line 381: Future improvement would be to be able to pass this as argument too. I know this is great for GH actions but not really that nice for local execution.

.github/workflows/weekly-summary.yml
```diff
@@ -0,0 +1,23 @@
+# Generate a weekly summary of test failures
+name: Weekly Summary
+on:
+  schedule:
+    # Run early Monday mornings, but long enough after daily test run for results
+    - cron: 0 8 * * 1
+  # be able to start this action manually from a actions tab when needed
+  workflow_dispatch:
+
```
jkonecny12, line 9: It would be great to specify here permissions.
It should be enough to have 

```
permissions:
  contents: read
```

scripts/weekly-summary
jkonecny12: In that case please name it. Also I wonder isn't it already defined in the python as a constant?

scripts/log2json
bcl: Ah, right, I dropped that from the json since it's all a flat list and easy enough to add up with when parsing the file.

scripts/weekly-summary
bcl: That's on purpose. I want to log problems, but not stop the process. That's also why it doesn't check the return code.

scripts/weekly-summary
bcl: passed is really close to 'pass' so I'll use success, since that's what it is looking for anyway.

scripts/weekly-summary
bcl: It's the missing date case from this:
`datetime.fromtimestamp(t.get("start_time", 0)).strftime("%Y%m%d")`

It's not a number, it's a string representation of what fromtimestamp returns when passed a 0. Basically a way to keep any entries without a timestamp collected together. I'll add a note to the fromtimestamp() line to make this clearer.

scripts/weekly-summary
bcl: You really do not want to pass tokens on the cmdline since it is visible to other users. Yes, not a problem in this specific case, but generally not a good idea. config files and environmental variables are the best way to set these.

jkonecny12 (approved):
Looks great to me.

bcl:
Thanks, someone will need to merge this, I don't have that permission in this repo.

PR #697, opened 2022-03-29
Add workflow to run daily kickstart tests via permian.

rvykydal:
his PR is replacing the draft https://github.com/rhinstaller/kickstart-tests/pull/691 of a GitHub workflow integrating Permian ([kickstart tests plugin](https://github.com/rhinstaller/permian/tree/main/libpermian/plugins/kickstart_test)) into our daily kickstart tests scenario runs. In the end it would replace https://github.com/rhinstaller/kickstart-tests/blob/master/.github/workflows/scenarios.yml.                     
An example run of the workflow on my fork: https://github.com/rvykydal/kickstart-tests/actions/runs/2054657771.
You can see there that after each test a summary/status message is added by Permian workflow (Permian workflow is parsing kickstart-test launcher output to report results of individual tests).

The related Permian kickstart-test plugin PR support was added in https://github.com/rhinstaller/permian/pull/14
The Permian plugin workflow currently [runs launcher](https://github.com/rvykydal/permian/blob/43e598f3bcd21fb463e27ad3164caa1b08b7112d/libpermian/plugins/kickstart_test/__init__.py#L349) locally. The parameters for the launcher come from:
- Test plans: Test case management is done with [tclib](https://github.com/rhinstaller/tclib). In the GitHub workflow test cases are generated from kickstart-tests repository and aggregated into scenario by test plans defining the tests to be run. The [test plans](https://github.com/rvykydal/kickstart-tests/tree/scenarios-permian-workflow-v4-with-url/testlib/test_plans) are generated from templates in the workflow using our [skip-testtypes](https://github.com/rhinstaller/kickstart-tests/blob/master/containers/runner/skip-testtypes) as the single point of truth. In the future the information will be stored in the test plans.
- [Setttings](https://github.com/rvykydal/permian/blob/kstests-daily-scenario-v2/libpermian/plugins/kickstart_test/settings.ini): Settings overrides are passed to permian via `-o` (`--overrides`) option.
- Event: selects the [test plan](https://github.com/rvykydal/kickstart-tests/tree/scenarios-permian-workflow-v4-with-url/testlib/test_plans) by its `type`, launcher `--platform` option by `platform`, and boot.iso and repositories location by `bootIso` and `kstestParams.urls` .

For example:

```
{
  'type' : 'github.scheduled.daily.kstest.rhel8',
  'bootIso': {
    'x84_64': 'file:///boot.updated.with.PR.iso'
  }
  'kstestParams' : {
    'platform' : 'rhel8'
    'urls': {
      x86_64' : {
        'installation_tree': 'http://example.org/the-rhel-9/compose/BaseOS/x86_64/os',
        'modular_url': 'http://example.org/the-rhel-9/compose/AppStream/x86_64/os',
        'ftp_url': 'ftp://example.org/the-rhel-9/compose/BaseOS/x86_64/os',
      }
    }
  }
}
```

Compared to the previous [draft](https://github.com/rhinstaller/kickstart-tests/pull/691) the `kstestParams.urls` is used to set both the boot.iso and the repository location in a single place.

I would propose to schedule the workflow daily on a separate kstest-permian runner to see how it works compared to our current workflow and then update regular kstest runners with permian dependencies (which are not big) https://gitlab.cee.redhat.com/rhinstaller/builders/-/merge_requests/107, turn off the current scenarios.yaml and switch this workflow to use the updated kstest runners.


.github/workflows/scenarios-permian.yml
```diff
@@ -117,6 +126,7 @@ jobs:
           ./pipeline --debug-log permian.log \
             -o library.directPath="${{ github.workspace }}/kickstart-tests/testlib" \
             -o workflows.dry_run=False \
+            -o kickstart_test.timeout="${{ steps.set_timeout.outputs.timeout }}" \
```
jkonecny12, line 144: Why are you not using the environment variable here directly? You could then drop the ``Set launcher timeout from job timeout`` step above.

jkonecny12 (approved):
Otherwise looks good to me.

.github/workflows/scenarios-permian.yml
rvykydal: Yes, became too complicated. I wanted to have the timeout 20m less then the workflow timeout (so that the tests timeout before the workflow and we have partial logs). Because I am not able to use the timeout-minutes directly I wanted to be explicit about the workaround in the step... I think I will just set the value of TIMEOUT_MINUTES to the value (460) with a comment and use it directly here.

jkonecny12: I'm not sure you understand me. The point is not to use the matrix timeout value but to use the `env:` created variable here directly without the middle step.

```
env:
 TIMEOUT_MINUTES: 480

...

- name: Run scenario ${{ matrix.scenario }} in container
....
  -o kickstart_test.timeout="$(($TIMEOUT_MINUTES-20))m" \
```

rvykydal: Yeah, but still, I'd do it more simple, see the commit message of https://github.com/rhinstaller/kickstart-tests/pull/697/commits/b5221032c76fa4ffb3ac3aa4afae30bbfec3c451I've added (may squash it).
I think I was using the step because I felt is wouldn't be so simple (would need to handle values less than 20 actually), but as I said, I'd do it simply now and use automation eventually in a separate PR.

.github/workflows/scenarios-permian.yml
```diff
@@ -20,6 +20,9 @@ jobs:
     env:
       TEST_JOBS: 16
       GITHUB_TOKEN: /home/github/github-token
+      # Can't figure out how to read the timeout-minutes value defined above
+      # so define it again as environment variable usable in run script
+      TIMEOUT_MINUTES: 480
```
VladimirSlavik: Maybe something like `${{ fromJSON(jobs.scenario.timeout-minutes) }}` - but advice like that is cheap, testing is the real expense...

.github/workflows/scenarios-permian.yml
```diff
@@ -0,0 +1,179 @@
...
+          set -eux
+          if [ "${{ matrix.scenario }}" == "daily-iso" ] || [ "${{ matrix.scenario }}" == "minimal" ]; then
+            echo "::set-output name=platform::fedora_rawhide"
+          elif [ "${{ matrix.scenario }}" == "rhel8" ]; then
+            echo "::set-output name=platform::rhel8"
+          elif [ "${{ matrix.scenario }}" == "rhel9" ]; then
+            echo "::set-output name=platform::rhel9"
+          else
+            echo "::set-output name=platform::"
+            echo "Scenario ${{ matrix.scenario }} can't be mapped to platform"
```
VladimirSlavik, line 93: Is this case rather a reason to fail?

.github/workflows/scenarios-permian.yml
```diff
@@ -0,0 +1,179 @@
...
+            kickstart-tests/data/logs/kstest-*/*.log
+            kickstart-tests/data/logs/kstest-*/anaconda/lorax-packages.log
+
+      - name: Collect Permian logs
+        if: always()
+        uses: actions/upload-artifact@v2
+        with:
+          name: 'logs-permian-${{ matrix.scenario }}'
+          path: |
+            permian/permian.log
```
VladimirSlavik, line 160: Does setting the xunit Permian output leave any "normal" logs?

scripts/generate-testcases.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/python3
+
+import argparse
+import os
+import sys
+import re
+from jinja2 import Template
```
VladimirSlavik, line 7: I assume this is installed on Permian runner hosts, as part of a still open PR for builders? There is so much that floats around our infra that it might make sense to comment that this bit is installed on the runners. I think it's an unusual place to have a dependency.

To see where this comes from, I checked the dockerfiles, install_dependencies, spec file, then realized this is the kickstart tests repo so checked other dockerfiles, and finally grepped half the disk... and didn't find any reference!

scripts/generate-testcases.py
```diff
@@ -0,0 +1,81 @@
...
+def get_tags_from_line(line):
+    m = re.match(r'^\s*(TESTTYPE=)(.*)', line)
+    if not m:
+        return None
+    _tt, tags = m.groups()
+    m = re.match(r'.*(TESTTYPE:-)(.*)', tags)
+    if m:
+        _tt, tags = m.groups()
+    tags = [tag.strip('"}') for tag in tags.split()]
+    return tags
```
VladimirSlavik, line 32: Ouch. It can't be helped right now, but... In long term, is it important to have these `TESTTYPE=${TESTTYPE:-"values"}` cases? This convoluted parsing is just the cherry on top of that.

VladimirSlavik (approved):
Thank you! Sorry for taking so long. I think this is fine, and probably should be merged and iterated upon.

VladimirSlavik (commented):
I also think the third commit "Share the code for fetching daily iso..." either had an accident during rebasing, or is not described too well, but never mind that either.

I don't want another round with this monster, it might win :-)

.github/workflows/scenarios-permian.yml
rvykydal: It is, your are right, I'll fix it.

.github/workflows/scenarios-permian.yml
```diff
@@ -0,0 +1,179 @@
...
+            echo "::set-output name=installation_tree::http://download.eng.bos.redhat.com/rhel-8/development/RHEL-8/latest-RHEL-8/compose/BaseOS/x86_64/os"
+            echo "::set-output name=modular_url::http://download.eng.bos.redhat.com/rhel-8/development/RHEL-8/latest-RHEL-8/compose/AppStream/x86_64/os"
+          elif [ "${{ matrix.scenario }}" == "rhel9" ]; then
+            echo "::set-output name=installation_tree::http://download.eng.bos.redhat.com/rhel-9/development/RHEL-9/latest-RHEL-9/compose/BaseOS/x86_64/os"
+            echo "::set-output name=modular_url::http://download.eng.bos.redhat.com/rhel-9/development/RHEL-9/latest-RHEL-9/compose/AppStream/x86_64/os"
+          else
+            echo "::set-output name=installation_tree::"
+            echo "::set-output name=modular_url::"
+            echo "Installation tree location for ${{ matrix.scenario }} not configured"
+          fi
```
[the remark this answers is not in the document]
rvykydal: Also good point, maybe we should check if boot_iso_from_scenario.outputs.boot_iso is defined here and fail if it is not.

scripts/generate-testcases.py
rvykydal: The reason is "inheritance" of .sh files for code deduplication (eg bindtomac-team-pre.sh including team-pre.sh). It is not great solution and it will go away if/when we migrate these metadata (test tags) into tclib testcases.

.github/workflows/scenarios-permian.yml
rvykydal: This log is created by the permian run (`--debug-log permian.log`).
AFAIK xunit reporter creates only .xml logs.

rvykydal:
> I also think the third commit "Share the code for fetching daily iso..." either had an accident during rebasing, or is not described too well, but never mind that either.

I thought I was addressing your suggestion here https://github.com/rhinstaller/kickstart-tests/pull/691#discussion_r827284239 by splitting out the commit.

scripts/generate-testcases.py
rvykydal: I guess it is indirect dependency of Permian (https://gitlab.cee.redhat.com/rhinstaller/builders/-/merge_requests/102), maybe even the original kickstart tests runner - I was able to import it on our regular kstest runner.

.github/workflows/scenarios-permian.yml
rvykydal: Oh yes, I tried quite hard to find a way, but for `jobs.scenario.timeout-minutes` `jobs` is undefined and for `job.scenario.timeout-minutes` the value is empty. Nevermind, it is a separate issue applicable also at our current solution.

.github/workflows/scenarios-permian.yml
rvykydal: Done

.github/workflows/scenarios-permian.yml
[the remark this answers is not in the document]
rvykydal: Done

.github/workflows/scenarios-permian.yml
```diff
@@ -0,0 +1,120 @@
...
+          path: kickstart-tests
+
+      - name: Generate test cases
+        working-directory: ./kickstart-tests
+        run: scripts/generate-testcases.py -t ./testlib/test_cases/kstest-template.tc.yaml.j2 . -o ./testlib/test_cases
+
+      - name: Clone Permian repositories
+        run: |
+          git clone https://github.com/rhinstaller/permian.git
+          git clone https://github.com/rhinstaller/tclib.git
```
jkonecny12, line 70: I would rather like to use checkout actions instead of clone.

.github/workflows/scenarios-permian.yml
```diff
@@ -0,0 +1,120 @@
...
+      - name: Set platform from scenario
+        id: platform_from_scenario
+        run: |
+          set -eux
+          if [ "${{ matrix.scenario }}" == "daily-iso" ] || [ "${{ matrix.scenario }}" == "minimal" ]; then
+            echo "::set-output name=platform::fedora_rawhide"
+          elif [ "${{ matrix.scenario }}" == "rhel8" ]; then
+            echo "::set-output name=platform::rhel8"
+          elif [ "${{ matrix.scenario }}" == "rhel9" ]; then
+            echo "::set-output name=platform::rhel9"
```
jkonecny12, line 90: I'm thinking if instead of these if's these shouldn't be specified by the matrix. That would simplify readability and future changes.

scripts/generate-testcases.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/python3
```
jkonecny12, line 1: I know it's missing on plenty of places but we should have a COPYLEFT header on the new files to start with that slowly.

testlib/test_cases/README.md
```diff
@@ -0,0 +1 @@
+This directory should contain test cases generated from kickstart test repository using scripts/generate-testcases.py script.
```
jkonecny12, line 1: I wonder, why we need to generate them and have them part of the repository. To me it seems like something which should be abstracted. One way we could go is to use their format and really have that as first-class citizen of KS tests.  Another way seems to me would be to not have these completely and generate them when required.

Honestly, I'm more inclined of start using them here as first-class citizen. For example we might migrate our TAGs to this solution and finally have a metadata for tests?

jkonecny12 (commented):
Looks good to me otherwise. However, I would like to discuss my points before giving ACK.

.github/workflows/scenarios-permian.yml
rvykydal: We may consider this as separate issue, but scenario (eg daily-iso / rawhide - configures the tested boot.iso) and platform (eg fedora_rawhide - configures the platform specific fragments ... --platform argument) seem to me as clearly different things.

testlib/test_cases/README.md
rvykydal: That is a follow-up goal I am planning to consider. But as a next isolated step. Moving the metadata will have consequences for various other usages so we need to be careful here,  for example: I want to run all network tests locally for which we need some tooling and easy local usage of permian/tclib (RTT is working on their packaging).  And yes, tags seem as the first step to me.
The test cases (test library with generated test plans and test cases) are part of repository now as a kind of placeholder for future migration.

.github/workflows/scenarios-permian.yml
VladimirSlavik: I cautiously agree, but only because of consistency.

I see no reason to consider the actions "better" or "correct". The whole workflow is de facto a shell script with some additional stuff instead of comments, so perhaps it's actually the actions that are out of place. After all, the only other actions we consistently use in rhinstaller org are related to the platform (api calls, upload files).

scripts/generate-testcases.py
rvykydal: Updated

.github/workflows/scenarios-permian.yml
rvykydal: Updated with the action.

rvykydal:
I've also updated the PR for Brian's changes related to weekly reports and rebased the PR on top of them (they can be seen in the changes to previous version.

PR #700, opened 2022-04-01
Support for kstest on pr with permian

rvykydal:
Support for https://github.com/rhinstaller/anaconda/pull/4000

### DIFF
--- a/scripts/generate-permian-query.py
+++ b/scripts/generate-permian-query.py
@@ -1,0 +1,40 @@
+#!/usr/bin/python3
+
+import argparse
+import itertools
+
+
+def parse_args():
+    _parser = argparse.ArgumentParser(
+        description="Generate tclib testcase query from kickstart tests launcher "
+                    "options --skip-testtypes, --testtype and a list of tests."
+    )
+    _parser.add_argument("--skip-testtypes", "-s", type=str, metavar="TYPE[,TYPE..]",
+                         nargs=1, action="append", help="skip tests with TYPE (tag)")
+    _parser.add_argument("--testtype", "-t", type=str, metavar="TYPE",
+                         help="only run tests with TYPE (tag)")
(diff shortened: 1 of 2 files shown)


scripts/generate-permian-query.py
```diff
@@ -0,0 +1,40 @@
...
+    args = parse_args()
+
+    conditions = []
+    if args.tests:
+        conditions = ['tc.name == "{}"'.format(test) for test in args.tests]
+    else:
+        if args.testtype:
+            conditions.extend(['"{}" in tc.tags'.format(args.testtype), '"knownfailure" not in tc.tags'])
+        if args.skip_testtypes:
+            skiptypes = ','.join(itertools.chain(*args.skip_testtypes)).split(',')
```
VladimirSlavik, line 32: You get a list of strings that contain comma-separated items, such as `["a,b", "c"]` and you want a list of all the items `["a", "b", "c"]` out of that? I tried that kind of input with this code in the interpreter and got a list of empty and one-letter strings, so I'm not really sure it works.

What worked for me was `itertools.chain(*map(lambda x: x.split(","), args.skip_testtypes))`. But I'm not really sure I tested the same thing as you did...

rvykydal: The input string list is rather [['a,b'], ['c']] here, for
```
scripts/generate-permian-query.py --skip-testtypes a,b --skip-testtypes c
```
and we are doing that because we add --skip-testtypes based on branch detection and --skip-testtypes from user request.

VladimirSlavik: Thank you, then I was wrong. Sorry for the noise!

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
LGTM

Issue #704, opened 2022-04-05
[weekly reports] Add summary also to daily results


rvykydal:
Quoting:

2) I think it could be useful to have such summary also for each day/scenario, eg
```
03/16/2022
    logs-daily-iso (SUMMARY):
```
It can became very useful for quick picture when there are mass failures, workflow timeouts, infra issues etc.

_Originally posted by @rvykydal in https://github.com/rhinstaller/kickstart-tests/pull/692#pullrequestreview-920265479_

bcl:
By summary do you mean the pass/fail count for each day?

eg.

```
  03/31/2022
    logs-daily-iso (Ran 247, 245 passed, 2 failed):
        bindtomac-bridge-httpks
        clearpart-2
        clearpart-2
    logs-rhel8 (Ran 194, 192 passed, 2 failed):
        network-device-mac
        ntp-without-chrony
```

rvykydal:
Yes.

bcl:
Fixed by #705 I guess t doesn't know about lists? :)

rvykydal:
Thank you!

PR #705, opened 2022-04-08
weekly-summary: Add more information about the tests

### LINKED ISSUE #702: [weekly reports] Take flakes into account in summary.
I opened the issue following on https://github.com/rhinstaller/kickstart-tests/pull/692#pullrequestreview-920265479, quoting:

I have some ideas regarding the summary (`logs-daily-iso: Ran 1534 tests. 1501 passed, 33 failed.`) :
1) Maybe I'd work with a concept of flake. We run the tests with --retry option which means in case of a test failure try to run the test again once more. We consider test that failed and then succeeded as a success (with flake). And I think this could be also visible in the summary. This is what I am using when creating current summary in permian (parsing the output / tests.log of the test run) (https://github.com/rvykydal/kickstart-tests/runs/5497126774?check_suite_focus=true, https://github.com/rhinstaller/permian/blob/339bcfda8269e25e40ad12db39ad2a812730a214/libpermian/plugins/kickstart_test/__init__.py#L116):
```
SUCCESS: 186 FAILED: 2 N/A: 33 (runs: 192 flakes: 2 timed out: 1)
```
`SUCCESS` - number of tests that passed (including flakes, ie fail,pass)
`FAILED` - number of tests that failed twice, ie excluding flake failures
`N/A` - tests for which we don't have a result yet
`runs` - number of all test runs (Ran: in your summary)
`flakes` - number of flakes (fail,pass of a test)
`timed out` - this info is also useful indicator sometimes

_Originally posted by @rvykydal in https://github.com/rhinstaller/kickstart-tests/pull/692#pullrequestreview-920265479_

bcl:
This adds a count of the 'flaky' tests, ones which fail at first but
eventually succeed. This now counts as a successful test, but is also
listed separately.

This adds an option to include detailed info about the flaky tests if
you pass `--flaky-details` to the script.

This now handles expired artifacts -- GitHub is not returning an error
if you try to retrieve an expired artifact, so it needs to handle trying
to unzip a non-zipfile (it's a json blob).

Remove unused functions, and split up some of the logic into functions.
Also improves documentation of the data structures created for
generating the reports and stats.

This should Fix #702, #704

### DIFF
--- a/scripts/weekly-summary
+++ b/scripts/weekly-summary
@@ -99,6 +99,10 @@ def get_artifacts(token, artifact_names, start, end):
 def run_curl(token, url, filename):
     """
     run_curl downloads an artifact file
+
+    It returns True if the response was a 200
+    If there was an exception is returns False and the error, as well as
+    printing it to stderr
     """
     with open(filename, "wb") as f:
         c = pycurl.Curl()
@@ -122,7 +126,8 @@ def run_curl(token, url, filename):
 
         try:
             c.perform()
-            ok = (True, None)
+            status = c.getinfo(pycurl.HTTP_CODE)
+            ok = (status == 200, None)
         except Exception as e:
             print(f"ERROR: {e}", file=sys.stderr)
             sys.stderr.flush()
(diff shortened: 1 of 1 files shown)


bcl:
You can view an example of the output here:
https://bcl.fedorapeople.org/logs/summary.log

rvykydal (approved):
Looks good to me, thank you.

bcl:
Yeah, sorry for the giant diff, I started out trying to do things in pieces, but didn't really know where I was going until the end :) I'm not sure how useful the flake details are, since they tend to always be timeouts. We can always turn them on/off later.

rvykydal:
> Yeah, sorry for the giant diff, I started out trying to do things in pieces, but didn't really know where I was going until the end :) I'm not sure how useful the flake details are, since they tend to always be timeouts. We can always turn them on/off later.

I think the details are useful, I'd keep them on by default.

PR #717, opened 2022-05-04
Use temporary artifact names for testing purposes

rvykydal:
Add -testing suffix so that the logs are not used in weekly reports at
the first stage of the deployment (running in parallel with non-permian
scenarios).


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #718, opened 2022-05-05
Schedule daily testing runs of daily kstests run via permian

rvykydal:
We will run them in parallel with current scenario for a while on a
dedicated separate kstests-permian runner.

Before merging:
- [x] Do not mess weekly reports with results:  https://github.com/rhinstaller/kickstart-tests/pull/717


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #721, opened 2022-05-06
Use permian branch with reverted regression

rvykydal:


poncovka (approved):

VladimirSlavik (approved):
If this is just for now, that seems the best option to get feedback from real runs.

rvykydal:
> If this is just for now, that seems the best option to get feedback from real runs.

Definitely temporary, we need to figure it out with RTT guys properly when I am back from PTO.

Issue #723, opened 2022-05-17
[weekly reports] 2 days missing in Weekly summary #7


rvykydal:
In Weekly summary #7 2 days (05/11/2022, 05/10/2022)  are missing:
https://github.com/rhinstaller/kickstart-tests/actions/runs/2330583161
I think it is caused by the fact that we are creating more artifacts now (testing running daily tests with Permian) and the dailies of the last 7 days don't fit in the page with artifacts fetched (using 100 artifacts per page which is the maximum)
https://github.com/rhinstaller/kickstart-tests/blob/15cbb19a958ee9f50338b361285b7391a6ff063b/scripts/weekly-summary#L41

rvykydal:
Thank you @bcl worked for me locally.

PR #726, opened 2022-05-17
weekly-summary: Add support for pages and rate limits

bcl:
The API query only returns 100 results at a time. All of the artifacts
may not be included in the first page.

This adds support for requesting all the pages and retrieving artifacts
within the date range and desired names from each page.

This also adds a check for the GitHub rate limits. If is is hit, and the
reset time is less than 90s in the future it will wait and try again. If
this fails 3 times it will then exit with an error.

### DIFF
--- a/scripts/weekly-summary
+++ b/scripts/weekly-summary
@@ -33,6 +33,7 @@ import shutil
 import sys
 from subprocess import check_call
 import tempfile
+import time
 from urllib.request import Request, urlopen
 import zipfile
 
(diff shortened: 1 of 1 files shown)


bcl:
This should fix #723

rvykydal (approved):
Looks good to me, thank you.

scripts/weekly-summary
```diff
@@ -75,20 +76,49 @@ def get_artifacts(token, artifact_names, start, end):
...
+            if r.status == 403:
+                try:
+                    reset = int(r.headers.get("X-RateLimit-Reset"))
+                except:
+                    raise RuntimeError("Hit GitHub rate limit. Reset header missing.")
+                if retry_limit == 0 or time.time() > reset or reset - time.time() > 90:
+                    raise RuntimeError("Hit GitHub rate limit. Reset is at %s" % time.ctime(reset))
+
+                # Try waiting until the reset time
+                time.sleep(reset - time.time())
```
VladimirSlavik, line 100: Maybe wait until a few moments later? Or is that practically guaranteed by the connections taking some time as well?

scripts/weekly-summary
```diff
@@ -75,20 +76,49 @@ def get_artifacts(token, artifact_names, start, end):
...
+        # Only include the artifacts within the date range and names
+        for a in data["artifacts"]:
+            if a["name"] not in artifact_names:
+                continue
+            updated_at = datetime.fromisoformat(a["updated_at"][:-1])
+            if start <= updated_at <= end:
+                artifacts.append(a)
+
+        if len(data["artifacts"]) < 100:
+            break
```
VladimirSlavik, line 118: Is this the right exit condition? Or rather, is this actually a sensible way to query the API? AIUI, the way it's written here, this tries to get all pages until last which is detected by having less than 100 results, or we keep going. So I see the following failure modes:
- If there are 32768 pages, we get 32768. I tried the query and we now have 37 pages that return something, starting at `2020-11-12`, whatever that means. Is there a way to limit this better? Surely not all 37 pages are needed to get the last week's data?
- If the last page is perfectly aligned by coincidence, which *will* happen one day, this breaks when trying to get the next nonexistent page. We get status 304 on things beyond end, which as coded dies due to `304 != 200`.

VladimirSlavik (changes requested):
Thank you, this looks almost good. However, I don't think this will work as-is in the long term.

VladimirSlavik (commented):
Forgot one more detail - We're de facto scraping the API here, so I think there should be also some wait while grabbing the pages, rate limit or not.

scripts/weekly-summary
bcl: You mean a few moments after the rest time? Yeah, I meant to do that...

scripts/weekly-summary
VladimirSlavik: Uh, it's not guaranteed to be ordered? That's nasty.

At any rate, hitting the limits very much depends if downloading the archives counts as API request or not.

The limit for authenticated api queries seems to be 5000 (couldn't find any exact number), and we'd be making the requests with the token, so we probably need to fit in that limit. 37 pages of 100 archives takes us to 3/4 of the limit already as-is. The "caching" in `download_artifacts` _would_ likely save most of the downloads, but the workflow runs on GH's cloud runners, so it's a new machine every time, with nothing cached.

Do feel free to dismiss my nack if you agree with Radek - I've stated my case, that is enough.

bcl: No, 37 pages should only count as 37 requests :) We only download the artifacts we need for the week, and this only runs once a week. Currently it's processing 3 artifacts for each day so it should be a max of 58 a week.

VladimirSlavik (approved):

PR #727, opened 2022-05-18
Permian update bootiso setting

rvykydal:
For the change in Permian instead of having empty bootIso 'x86_64' value we don't define the bootIso structure at all so that the conversion from kstestParams.url can take place.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #728, opened 2022-05-18
Fix the selection of the Fedora IoT profile

poncovka:
Use inst.profile instead of inst.product and inst.variant.


rvykydal (approved):
Looks good to me.

PR #741, opened 2022-06-17
Disable fedora-live-image-build test temporarily (gh#740)

rvykydal:
Disable fedora-live-image-build until gh#740 is fixed.


VladimirSlavik (approved):
Thanks!

PR #744, opened 2022-06-22
Enable rootpw-allow-ssh test on rhel-9 (rhbz#2033849)

jstodola:
The feature is included in RHEL-9.1 development composes.


rvykydal (approved):
Thank you

PR #754, opened 2022-07-21
Disable network-addr-gen-mode-* tests on rhel8 temporarily (gh#748)

rvykydal:
Disable network-addr-gen-mode-* tests on rhel8 until gh#748 is fixed
(a new NM build gets in).


rvykydal:
It seems that passing the new NM build through gating on RHEL 8 can take some more time so disable the test until it happens.

VladimirSlavik (approved):
Thank you!

PR #755, opened 2022-07-27
Increase timeout for daily scenarios by 30 minutes

rvykydal:
We are tight to the limit for daily-iso and have some instances of 1
missing test in the run. I will look also if refreshing the runners
helps.


poncovka (approved):
Thank you!

VladimirSlavik (approved):
Thanks!

PR #762, opened 2022-08-15
Ignore "Call Trace:" in logs temporarily

rvykydal:
Ignore Call Traces so that gh761 doesn't block Anaconda testing and
development of testing tools.


jkonecny12 (approved):
Thanks for enabling the tests for us again.

VladimirSlavik (approved):
Thank you!

PR #763, opened 2022-08-15
Add the missing reqpart command

poncovka:
Add the `reqpart` command to all tests with custom partitioning. It will ensure
creation of the `biosboot` partition if required by the installed system.

### DIFF
--- a/btrfs-1.ks.in
+++ b/btrfs-1.ks.in
@@ -6,6 +6,7 @@ bootloader --timeout=1
 zerombr
 clearpart --all
 
+reqpart
 part --fstype=ext4 --size=500 /boot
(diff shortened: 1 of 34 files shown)


poncovka:
Some of the tests might require additional changes in the `%post` scripts. I didn't run all of them, so I suggest to merge this and open another pull request if necessary.

jkonecny12:
I wonder is this a behavior change of Anaconda or a side effect?

rvykydal (approved):
Thank you.

jkonecny12 (approved):
Thanks, my question was already answered.

The issue is that we have switched to `GPT` by default recently (there is a Fedora change about that).

PR #765, opened 2022-08-15
Add tests for the inst.disklabel boot option

poncovka:
Check the type of the partition table created during the installation.

### DIFF
--- a/disklabel-default.ks.in
+++ b/disklabel-default.ks.in
@@ -1,0 +1,21 @@
+#test name: disklabel-default
+
+# Use defaults.
+%ksappend repos/default.ks
+%ksappend common/common_no_payload.ks
+%ksappend payload/default_packages.ks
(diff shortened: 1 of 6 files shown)


rvykydal (approved):
Looks good to me, thank you.

jkonecny12 (approved):
LGTM!

poncovka:
Updated. All these tests should be skipped on RHEL 8 and RHEL 9 due to the unsupported `inst.disklabel` boot option.

rvykydal:
I think it would be great to have an option to run our daily scenarios (ie rhel8, rhel9, daily-iso) with selected tests on a PR like this.

rvykydal:
@poncovka I guess the build failed because of rawhide not having the required changes. They should be in now but I am not able to re-run the check.

PR #766, opened 2022-08-16
Add reqpart to raid 1

rvykydal:
The test was missisng in the https://github.com/rhinstaller/kickstart-tests/pull/763

### DIFF
--- a/raid-1.ks.in
+++ b/raid-1.ks.in
@@ -7,12 +7,12 @@ bootloader --timeout=1
 clearpart --all --initlabel
 
 reqpart
-partition raid.01 --size=500 --ondisk=vda --asprimary
-partition raid.02 --size=500 --ondisk=vdb --asprimary
-partition raid.11 --size=4000 --ondisk=vda
-partition raid.12 --size=4000 --ondisk=vdb
-partition raid.21 --size=1024 --ondisk=vda
-partition raid.22 --size=1024 --ondisk=vdb
+part raid.01 --size=500 --ondisk=vda --asprimary
+part raid.02 --size=500 --ondisk=vdb --asprimary
+part raid.11 --size=4000 --ondisk=vda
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
Looks good to me. Thanks!

poncovka (approved):

rvykydal:
Well, the change passes on rawhide, but on daily-iso I am getting
```
09:43:12,910 WARNING org.fedoraproject.Anaconda.Modules.Storage:INFO:program:Installing for i386-pc platform.
09:43:12,911 WARNING org.fedoraproject.Anaconda.Modules.Storage:INFO:program:grub2-install: warning: ../grub-core/partmap/gpt.c:194:this GPT partition label contains no BIOS Boot Partition; embedding won't be possible.
09:43:12,911 WARNING org.fedoraproject.Anaconda.Modules.Storage:INFO:program:grub2-install: error: embedding is not possible, but this is required for RAID and LVM install.
09:43:12,912 WARNING org.fedoraproject.Anaconda.Modules.Storage:DEBUG:program:Return code: 1
[Uploading virt-install.log…]()

```
 and a Traceback later

poncovka:
I guess the issue is that `reqpart` creates the `biosboot` partition only on one disk. In that case, it is a bug in Anaconda. We could use something like this as a workaround:

```
part biosboot --fstype=biosboot --size=1 --ondisk=vda
part biosboot --fstype=biosboot --size=1 --ondisk=vdb
```

poncovka:
@jstodola @vojtechtrefny Do you think `reqpart` should create both `biosboot` partitions by default? Otherwise, we could just add a new storage validation check that would block the installation in this case.

jstodola:
IMO, `reqpart` should create as many `biosboot` partitions as necessary. If two `biosboot` partitions are needed, they I would expect that the `reqpart` command would create them both.

PR #771, opened 2022-08-17
Ignore kernel Call Trace in log monitor (gh#768)

rvykydal:
Usually the Call Trace is not fatal for installation but log monitor
makes all the tests in the run fail. The reports for kernel team don't
seem to be useful either.


poncovka (approved):

VladimirSlavik (approved):
Thanks!

Issue #772, opened 2022-08-17
autopart-encrypted-1 fails on rhel8 with "lsblk: unknown column: mountpoints"


rvykydal:
lsblk on RHEL8 does not support "MOUNTPOINTS" --output (just "MOUNTPOINT" but I am not sure if that would work for what we are looking for).

```
00:51:46,756 INFO anaconda:program: Running in chroot '/mnt/sysroot'... /bin/sh /tmp/ks-script-fz8lwxg6
00:51:46,837 INFO anaconda:program: NAME                                          MAJ:MIN RM   SIZE RO TYPE  MOUNTPOINT
00:51:46,837 INFO anaconda:program: loop0                                           7:0    0 747.1M  1 loop
00:51:46,837 INFO anaconda:program: loop1                                           7:1    0     4G  1 loop
00:51:46,838 INFO anaconda:program: |-live-rw                                     253:0    0     4G  0 dm
00:51:46,838 INFO anaconda:program: `-live-base                                   253:1    0     4G  1 dm
00:51:46,838 INFO anaconda:program: loop2                                           7:2    0    32G  0 loop
00:51:46,839 INFO anaconda:program: `-live-rw                                     253:0    0     4G  0 dm
00:51:46,839 INFO anaconda:program: sr0                                            11:0    1   859M  0 rom
00:51:46,839 INFO anaconda:program: sr1                                            11:1    1   859M  0 rom
00:51:46,840 INFO anaconda:program: zram0                                         251:0    0   1.9G  0 disk  [SWAP]
00:51:46,840 INFO anaconda:program: vda                                           252:0    0    10G  0 disk
00:51:46,840 INFO anaconda:program: |-vda1                                        252:1    0     1G  0 part  /boot
00:51:46,841 INFO anaconda:program: `-vda2                                        252:2    0     9G  0 part
00:51:46,841 INFO anaconda:program: `-luks-a02c85d8-8953-4c60-aff7-d76bde330903 253:2    0     9G  0 crypt
00:51:46,841 INFO anaconda:program: |-rhel_ibm--p8--kvm--03--guest--02-swap   253:3    0     1G  0 lvm   [SWAP]
00:51:46,841 INFO anaconda:program: `-rhel_ibm--p8--kvm--03--guest--02-root   253:4    0     8G  0 lvm   /
00:51:46,842 INFO anaconda:program: lsblk: unknown column: mountpoints
00:51:46,844 DEBUG anaconda:program: Return code: 1
[...]

PR #776, opened 2022-08-17
Use the mountpoint column of the lsblk command

poncovka:
It looks like `lsblk` doesn't support `--output mountpoints` on RHEL 8.
Let's use the `mountpoint` column instead. The root device shouldn't be
mounted anywhere else in chroot.

Resolves: https://github.com/rhinstaller/kickstart-tests/issues/772


VladimirSlavik (approved):
Thanks!

rvykydal (approved):
Thank you

PR #785, opened 2022-09-12
ci(lint): add shell linter - Differential ShellCheck

jamacku:
Differential ShellCheck is a GitHub action that performs **differential ShellCheck scans** on shell scripts changed via PR and reports results directly in PR. The beauty of differential scans is that you will get **reports only about newly added defects**. 

Since this repository has many shell scripts, I think you might find some value in having a shell linter. Differential ShellCheck action is able to produce reports in SARIF format. GitHub understands this format and is able to display it nicely as a PR comment, and on the `Files Changed` tab, please see below.

image

image

Documentation is available at [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting. I'm always happy to extend functionality.

### DIFF
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -1,0 +1,26 @@
+# Doc: https://github.com/redhat-plumbers-in-action/differential-shellcheck#usage
+---
+
+name: Differential ShellCheck
+on:
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Differential ShellCheck
+        uses: redhat-plumbers-in-action/differential-shellcheck@v3
+        with:
+          severity: warning
+          token: ${{ secrets.GITHUB_TOKEN }}


VladimirSlavik:
Thank you, while I can't say we will take this, it seems to be a very useful tool. I assume this applies `.shellcheckrc` files as they are found, because it uses the cli linter?

jamacku:
Yes, it takes into consideration `.shellcheckrc` if it is defined in the root directory of your project.

VladimirSlavik:
(This might be interesting for our main repo, too. However, I'm not sure how to mix that with having shellcheck as an automake test at the same time.)

jamacku:
> (This might be interesting for our main repo, too. However, I'm not sure how to mix that with having shellcheck as an automake test at the same time.)

It should not be an issue, I expect that when automake shellcheck test fails contributors have to dig into logs to locate the shellcheck report. By introducing also `differential-shellcheck` you will also get a failed check and at the same time comment on PR describing an issue and exact location. I think having both will complement each other and improve the user experience of contributors.

jkonecny12 (commented):
I like this idea. Let's discuss this in the team.

jamacku:
Thank you. I can also prepare PR for the anaconda repo if you are interested.

jkonecny12:
The question about Anaconda repository is, if we can specify just a subset of the repository what should be checked. Even though, not sure if we need that or not?

VladimirSlavik:
For the main repo, the issue I see is that we have also `.shellcheckrc` in subdirectories. If this action works the way I think, it should be fine and result in the same outputs as the autotools-backed shellcheck "unit test". But I am not sure. But, maybe for a differential check, a divergent config would be good enough? I mean, it's the same tool so surely it won't contradict itself...

Either way, if you want to make a PR and save me some thinking, I'd be grateful!

jamacku:
@VladimirSlavik I will prepare PR for the anaconda repo, test behavior, and check if it is the same as your autotools ShellCheck.

PR #793, opened 2022-09-16
Enable RHEL tests that require a DVD ISO

poncovka:
Remove the `server-dvd-link` fragment. Use the `KSTEST_DVD_ISO_URL`
variable instead. The URL of the ISO can contain regexes in the ISO name.
For example: `http://my-server/Fedora-.*?.iso`

Tests that fail to download the required DVD ISO should fail early.
Use the `--erroronfail` option to achieve that.

Set higher timeout for tests that download a DVD ISO.

**TESTING IN PROGRESS**


poncovka:
Unfortunately, this solution is unusable for RHEL, because the DVD ISO is too big. We will try to use a fake ISO instead.

PR #809, opened 2022-10-11
daily scenario: increase XOrg timeout by 40 seconds (gh#795)

rvykydal:
Address the kickstart-test issue gh#795 to reduce rising number of flakes on XOrg timeout.


VladimirSlavik (approved):
Thank you!

PR #813, opened 2022-10-17
Disable groups-and-envs-2 on daily-iso temporarrily (gh#812)

rvykydal:
Disable groups-and-envs-2 on daily-iso until gh#812 is fixed.


VladimirSlavik (approved):
Thanks!

PR #820, opened 2022-11-01
Fix bond-ks-initramfs on RHEL 8

rvykydal:


VladimirSlavik (approved):
Thank you, looks good to me.

PR #836, opened 2022-12-06
Update default RHEL repositories to 8.8 and 9.2

rvykydal:
This is for local runs using rhel isos and --platform rhelX.
For daily workflow it has been changed here: https://github.com/rhinstaller/kickstart-tests/pull/796


VladimirSlavik (approved):
Thanks!

PR #215, opened 2019-01-22
Add dry run mode

jkonecny12:
This will be helpful for development and also for implementation of unit-tests.

PR #838, opened 2022-12-06
Update network tests counting NM connections for new NM version (#215…

rvykydal:
…0773)

The loopback device is now managed with lo connection created in initramfs.

Tracked in gh#835.


poncovka (commented):
Thanks!

PR #850, opened 2023-01-03
Test installation with a preexisting Btrfs

peberanek:
This test covers https://bugzilla.redhat.com/show_bug.cgi?id=2139169 and https://bugzilla.redhat.com/show_bug.cgi?id=2139166


preexisting-btrfs.sh
```diff
@@ -0,0 +1,43 @@
...
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+# Red Hat Author(s): Petr Beranek <pberanek@redhat.com>
+
+# This test covers https://bugzilla.redhat.com/show_bug.cgi?id=2139169 and
+# https://bugzilla.redhat.com/show_bug.cgi?id=2139166
```
VladimirSlavik, line 22: I think there should be a word or two about what this does, so we don't have to go read the bugs next time too? ;-)

preexisting-btrfs.sh
```diff
@@ -0,0 +1,43 @@
...
+#   sudo losetup -f disk-a.img
+#   sudo parted --script /dev/<device> mklabel gpt  # use 'sudo losetup -l' to
+#                                                   # discover actual <device> name
+#   sudo parted --script /dev/<device> mkpart p1 btrfs 0% 100%
+#   sudo mkfs.btrfs /dev/<device><partition>  # use 'lsblk' to discover actual
+#                                             # <partition> name
+#   sudo losetup -d /dev/<device>
+#   tar caSf preexisting-btrfs.tar.xz disk-a.img
+
+TESTTYPE="btrfs storage"
```
VladimirSlavik, line 36: Please add a comment explaining why we want to run the test on rhel too. The first thought about this place would be usually: btrfs+rhel=nope so why is `skip-on-rhel` missing here? Something like "enabled on rhel since it must be able to install over btrfs", perhaps...

VladimirSlavik (approved):
Thank you! Looks good to me, except for the few missing comments below. Having a few words already in the files would go a long way towards making this easy to understand in the future.

VladimirSlavik:
Thanks for fixing the shellcheck thing. However, perhaps we should take a systematic approach to that, across the whole repo :)

peberanek:
@VladimirSlavik Thanks for comments. Sure, I'm gonna to add some comments to provide more context about the bug.

Concerning shellcheck: I definitely support a systematic approach (I'm not a shellcheck expert, unfortunately). Is it ok to leave the `shellcheck disable` directives in the code now, or do you plan some changes or tweaks in the shellcheck test in the near future?

rvykydal (approved):
Looks good to me, thank you!

rvykydal:
Should I create an issue to handle the shellcheck issue on configuration level ?

rvykydal:
> Should I create an issue to handle the shellcheck thing on configuration level ? For me it is ok to leave the directives in this PR.

https://github.com/rhinstaller/kickstart-tests/issues/853

peberanek:
I think the problem are container privileges as `parted` and `mkfs.btrfs` needs to be run as root. Correct? A similar approach is used here: https://github.com/rhinstaller/kickstart-tests/pull/781#issuecomment-1234240273 But if there is a simpler or safer solution, let's do it :-)

rvykydal:
> Couldn't we somehow easily generate the tar.xz archive? It's a bit problematic for me that it could contain basically anything.

I am not sure if it is better approach - makes the test less deterministic? IIRC we've been discussing it with @poncovka.

jstodola:
Another advantage of pre-built images is that they can be created by any version of tools. For example a file system can be formatted by `mkfs.xfs` from RHEL-7.

peberanek:
Hi, any updates to his PR? If there are no objections, I believe it is ready for merge.

rvykydal:
I agree.

PR #851, opened 2023-01-03
Add test for stage2 from ks

rvykydal:


functions.sh
```diff
@@ -402,3 +402,14 @@ append_additional_repo_to_kernel_args() {
 get_timeout() {
     echo "60"
 }
+
+stage2_from_ks() {
+    # Return true to get stage2 location from kickstart (inst.stage2 option will not be supplied).
```
poncovka, line 407: There can be more use cases that don't rely on the `inst.stage2` option. For example, it could be specified with `inst.repo`. It would be great if the new support could be eventually used to test all scenarios, but it is just a suggestion.

poncovka (approved):
Thank you!

functions.sh
rvykydal: https://github.com/rhinstaller/kickstart-tests/issues/873

PR #855, opened 2023-01-06
Disable shellcheck SC2034

rvykydal:
It is failing on our current tags definition via TESTTYPE variable which is actually parsed out by toolong scripts.


VladimirSlavik:
I wonder if this could be made more specific just for `TESTTYPE`. It seems that no :( In that case it's probably fine.

jstodola:
Maybe something like:
`# Ignore unused variable parsed out by tooling scripts`
`# shellcheck disable=SC2034`

before each `TESTTYPE=...` line?

rvykydal:
> Maybe something like: `# Ignore unused variable parsed out by tooling scripts` `# shellcheck disable=SC2034`
> 
> before each `TESTTYPE=...` line?

That would scratch the whole effort following https://github.com/rhinstaller/kickstart-tests/pull/850#issuecomment-1369797480 :). But I am not against that - ie add #/bin/bash to all .sh files, add a comment and disable locally for all .sh files (like in https://github.com/rhinstaller/kickstart-tests/pull/850).
I think root problem here is using TESTTYPE the way we do (parsing as test metadata with sed in kstests tooling). This will go away when we migrate to real metadata (like in https://issues.redhat.com/browse/INSTALLER-3109)

rvykydal:
> Maybe something like: `# Ignore unused variable parsed out by tooling scripts` `# shellcheck disable=SC2034`
> 
> before each `TESTTYPE=...` line?

We decided to go with this solution.

Issue #853, opened 2023-01-04
Prevent shellcheck warnings for missing shebang and TESTTYPE definition.


rvykydal:
For example as in this PR: https://github.com/rhinstaller/kickstart-tests/pull/850
where it was handled by adding the shebang and disabling the checks in the test.
We need a general solution for these warnings.

VladimirSlavik:
Is adding the shebangs ok?

rvykydal:
> Is adding the shebangs ok?
> 
> I expect that all the actual results would start pouring out after this.

I think so, let's add them. Or we can use "SHELLCHECK_OPTS -s bash" as in https://github.com/rhinstaller/kickstart-tests/pull/855

rvykydal:
For the TESTTYPE we agreed on adding a comment and disabling locally for each test. I'd suggest:
https://github.com/rvykydal/kickstart-tests/blob/f1348ef2602317a26a2b89cd41a3c594dd251ccb/stage2-from-ks.sh#L20

rvykydal:
> For the TESTTYPE we agreed on adding a comment and disabling locally for each test. I'd suggest: https://github.com/rvykydal/kickstart-tests/blob/f1348ef2602317a26a2b89cd41a3c594dd251ccb/stage2-from-ks.sh#L20

https://github.com/rhinstaller/kickstart-tests/pull/862

PR #865, opened 2023-01-12
Make preexisting-btrfs consistent with other tests wrt shellcheck

rvykydal:
Use the same solution as we did for all tests in
https://github.com/rhinstaller/kickstart-tests/issues/853


poncovka (approved):
Thanks!

VladimirSlavik (approved):
Thanks!

PR #877, opened 2023-01-27
Disable network-bootopts-noautodefault on daily-iso temporarily (gh#874)

rvykydal:
Disable network-bootopts-noautodefault on daily-iso until gh#874 is fixed.


VladimirSlavik (approved):
Thanks!

PR #896, opened 2023-02-27
Fix hibernation test failing when UUID is used

OndrejZobal:
autopart-hibernation test will now use both path to block device and UUID to verify that hibernation was set up correctly.

When I was running Kickstart tests locally with a "rawhide" image `/dev/mapper/.*-swap` was always used. On daily ISOs UUID is used. Not sure why but both should work now!


rvykydal (approved):
Looks good to me, thank you.

PR #901, opened 2023-03-06
Update encrypt-devices test (gh#898)

rvykydal:
On RHEL 8 with the default package selection, DNF started to require additional space on /var.

"At least 40MB more space needed on the /var filesystem."


poncovka (approved):
Thank you!

PR #919, opened 2023-03-27
Disable lvm-1 temporarily on daily-iso (gh#918)

rvykydal:
Disable lvm-1 on daily-iso until gh#918 is fixed.


VladimirSlavik (approved):
Thanks!

Issue #918, opened 2023-03-27
lvm-1 failing on daily-iso with "*** root lv did not preserve --fsoptions" 


rvykydal:
Started to appear on 03-24-2023
```
*** root lv did not preserve --fsoptions
```
[kstest.log](https://github.com/rhinstaller/kickstart-tests/files/11078809/kstest.log)
[virt-install.log](https://github.com/rhinstaller/kickstart-tests/files/11078810/virt-install.log)

rvykydal:
packages diff:
```
< NetworkManager-1.43.3-1.fc39.x86_64
< NetworkManager-libnm-1.43.3-1.fc39.x86_64
< NetworkManager-team-1.43.3-1.fc39.x86_64
< NetworkManager-wifi-1.43.3-1.fc39.x86_64
---
> NetworkManager-1.43.4-1.fc39.x86_64
> NetworkManager-libnm-1.43.4-1.fc39.x86_64
> NetworkManager-team-1.43.4-1.fc39.x86_64
> NetworkManager-wifi-1.43.4-1.fc39.x86_64
12,18c12,18
< anaconda-39.6-1.20230321164102073103.master.16.g7c035ed2ad.fc39.x86_64
< anaconda-core-39.6-1.20230321164102073103.master.16.g7c035ed2ad.fc39.x86_64
< anaconda-dracut-39.6-1.20230321164102073103.master.16.g7c035ed2ad.fc39.x86_64
< anaconda-gui-39.6-1.20230321164102073103.master.16.g7c035ed2ad.fc39.x86_64
< anaconda-install-env-deps-39.6-1.20230321164102073103.master.16.g7c035ed2ad.fcc
39.x86_64
< anaconda-install-img-deps-39.6-1.20230321164102073103.master.16.g7c035ed2ad.fcc
39.x86_64
< anaconda-tui-39.6-1.20230321164102073103.master.16.g7c035ed2ad.fc39.x86_64
---
> anaconda-39.6-1.20230323165550131553.master.34.gd1f26a2e33.fc39.x86_64
> anaconda-core-39.6-1.20230323165550131553.master.34.gd1f26a2e33.fc39.x86_64
> anaconda-dracut-39.6-1.20230323165550131553.master.34.gd1f26a2e33.fc39.x86_64
> anaconda-gui-39.6-1.20230323165550131553.master.34.gd1f26a2e33.fc39.x86_64
[...]

PR #921, opened 2023-03-29
Use both UUID and path for checking fsoptions presence in fstab

vojtechtrefny:
Blivet recently started using UUID instead of path for LVM (it already uses UUID for other devices) so the check for the options presence needs to be adjusted.

Fixes: #918


rvykydal (approved):
Looks good to me, thank you!

Issue #920, opened 2023-03-29
In daily-iso scenario log Udisks error appears: "/usr/libexec/udisks2/udisksd: undefined symbol: bd_loop_get_autoclear"


rvykydal:
Started to appear: 03-24-2023

```
00:13:57,636 INFO systemd:Starting udisks2.service - Disk Manager...
00:13:57,668 INFO udisksd:/usr/libexec/udisks2/udisksd: symbol lookup error: /usr/libexec/udisks2/udisksd: undefined symbol: bd_loop_get_autoclear
00:13:57,678 NOTICE systemd:udisks2.service: Main process exited, code=exited, status=127/n/a
00:13:57,678 WARNING systemd:udisks2.service: Failed with result 'exit-code'.
00:13:57,681 ERR systemd:Failed to start udisks2.service - Disk Manager.
00:13:57,683 WARNING org.fedoraproject.Anaconda.Modules.Storage:INFO:blivet:failed to get initiator name from iscsi firmware: UDisks iSCSI functionality not available
```

[kstest.log](https://github.com/rhinstaller/kickstart-tests/files/11097723/kstest.log)
[virt-install.log](https://github.com/rhinstaller/kickstart-tests/files/11097724/virt-install.log)

Packages diff:
[diff.package_versions.log](https://github.com/rhinstaller/kickstart-tests/files/11097757/diff.package_versions.log)

rvykydal:
@vojtechtrefny Hi Vojta, are you aware of this issue ?

vojtechtrefny:
We are doing some API changes in libblockdev (C only, everything in Python should still be backwards compatible) so you'll need the daily builds of UDisks too to make sure it still works. We should release 3.0 soon so these things shouldn't happen again (but we are saying "soon" for two years now).

PR #922, opened 2023-03-30
Enable @storage/udisks-daily repo for the daily ISO workflow

vojtechtrefny:
The blivet-daily Copr repository also contains the daily builds of libblockdev which has some API changes for the upcoming 3.0 release so we need the daily builds of UDisks too to make sure it works.

Fixes: #920


rvykydal (approved):
Looks good to me, thank you.

PR #945, opened 2023-05-09
Disable rpm-ostree temporarily on daily-iso (gh#942)

rvykydal:
Disable rpm-ostree test on daily-iso until gh942 is fixed.


VladimirSlavik (approved):
Thanks!

PR #958, opened 2023-05-30
Fix the return value of validate (gh#957)

rvykydal:
The validate function should return a status, not a message.

Application of commit d1cd47d31de588746a6b8985891b585786a41ef1 to other relevant occurences.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #965, opened 2023-06-07
Reduce the timeout for copying files from installed image (gh#964)

rvykydal:
300s should be enough, let's not block the resources longer than needed. Also the change may put some light on related gh#964.


VladimirSlavik (approved):
Yes, please.

PR #968, opened 2023-06-13
Enable network-dns-search on RHEL

jstodola:
This feature is implemented in RHEL-8.9 (bug 1656662) and RHEL-9.3 (bug 2065754).


VladimirSlavik (approved):
Thanks!

PR #971, opened 2023-06-13
Disable selinux-contexts temporarily on rhel9 (gh#970)

rvykydal:
Disable selinux-contexts test on rhel-9 until gh970 is fixed.


VladimirSlavik (approved):
Thanks!

PR #972, opened 2023-06-13
Disable raid-ddf temporarily on rhel (gh#969)

rvykydal:
Disable raid-ddf test on rhel until gh969 is fixed.


VladimirSlavik (approved):
Thanks!

PR #976, opened 2023-06-22
Disable packages-default temporarily on daily-iso (gh#975)

rvykydal:
Disable packages-default test on daily-iso until gh975 is fixed.


VladimirSlavik (approved):
Thanks!

Issue #942, opened 2023-05-09
rpm-ostree failing on daily-iso: "*** The test has failed."


rvykydal:
05-03-2023
```
2023-05-03 04:30:59,007 DEBUG: Failed to run subprocess: '['/opt/kstest/kickstart-tests/scripts/launcher/lib/launcher_interface.sh', '-i', '/opt/kstest/data/images/boot.iso', '-k', '1', '-w', '/var/tmp/kstest-rpm-ostree.2023_05_03-04_21_25._c7rq7mc', '-t', '/opt/kstest/kickstart-tests/rpm-ostree.sh', 'validate']'
stdout:
*** The test has failed.

2023-05-03 04:30:59,008 INFO: RESULT:rpm-ostree:719881f911af:FAILED:Validation failed with return code 1
```
```
3:46:24,743 INFO anaconda:program: Unexpected URL: https://ostree.fedoraproject.org/iot
```

[virt-install.log](https://github.com/rhinstaller/kickstart-tests/files/11430651/virt-install.log)
[Uploading kstest.log…]()

rvykydal:
PR to disable the test: https://github.com/rhinstaller/kickstart-tests/pull/945

rvykydal:
@jkonecny12 maybe you are aware of some related change.

jkonecny12:
Unfortunately not, but I'll try to find out.

jkonecny12:
So I was looking into this today. Seems that the issue is that the koji link we have in the test works but changing url after installation to `https://ostree.fedoraproject.org/iot`. The test fails but it seems (wasn't able to verify) that it is correct.

However, I looked on possibility to use this official link instead and that is failing. Reason is broken commits in the ostree repository which is filed here https://pagure.io/fedora-infrastructure/issue/11317 .

I would give a few days if there will be resolution for the broken ostree and use the official repository instead of koji. If that won't work than we can just change the test to verify against the new url.

PR #977, opened 2023-06-24
Enable rpm-ostree test again

jkonecny12:
It's working again now. Probably because of changes done by Fedora rel-engs when trying to resolve this:

https://pagure.io/fedora-infrastructure/issue/11317

Closes https://github.com/rhinstaller/kickstart-tests/issues/942


VladimirSlavik (approved):
Thanks! I see that the test passed now, so it's ok.

rvykydal (approved):
Thank you.

Issue #1000, opened 2023-09-06
Fix manual *-luks-* test


rvykydal:
The *-luks-* tests are currently failing on rawhide and will be failing (in a bit different way) when https://github.com/rhinstaller/anaconda/pull/5081 is merged.
We should update them so we can use them.

rvykydal:
PR with fix: https://github.com/rhinstaller/kickstart-tests/pull/1002

PR #1002, opened 2023-09-06
Update some manual storage tests for new minimal ESP size (500MB)

rvykydal:
Related to this change:
https://github.com/rhinstaller/anaconda/commit/c253390b23475970f1d696b4fb9d719a0c316559

Minimum size increased by 300MB.

The reason why the tests are manual (and therefore rot a bit not being run daily) are that libvirt has troubles getting the RESULT from the installer image, therefore manual check is needed.

From my local testing seems we can make part-luks-* automatic, I'll follow on it.
I've checked the patch locally, I guess the CI will fail, maybe except for part-luks-*.

Should fix: https://github.com/rhinstaller/kickstart-tests/issues/1000

### DIFF
--- a/lvm-luks-1.ks.in
+++ b/lvm-luks-1.ks.in
@@ -12,7 +12,7 @@ clearpart --all --initlabel
 
 reqpart
 part /boot --fstype="ext4" --size=1024
-part pv.1 --fstype="lvmpv" --size=9215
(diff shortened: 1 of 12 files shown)


VladimirSlavik (commented):
Thank you!

VladimirSlavik (approved):
Thank you!

PR #1004, opened 2023-09-07
Enable part-luks-* tests on fedora.

rvykydal:
Reading of tests results on Fedora seems to work reliably enough. On RHEL it is not the case.

Depends on https://github.com/rhinstaller/kickstart-tests/pull/1002


VladimirSlavik (approved):
Thanks!

PR #1007, opened 2023-09-12
infra: Raise kstest errors in the permian step

rvykydal:
This makes it so that opening the GH web page automatically opens logs with the actual errors. Previously, it showed only a cryptic message "Runner return code: 1". and the actual errors were in a successfully completed step.

I've applied @VladimirSlavik's  https://github.com/rhinstaller/anaconda/commit/7120a788d0216e71305e5bd9da0940ed416c0c92 also here.

### DIFF
--- a/.github/workflows/scenarios-permian.yml
+++ b/.github/workflows/scenarios-permian.yml
@@ -179,6 +179,14 @@ jobs:
               }
             }'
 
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Thanks!

PR #1008, opened 2023-09-20
Add script scanning weekly report logs for existing issues

rvykydal:


scripts/classify-failures
```diff
@@ -0,0 +1,254 @@
...
+import sys
+
+ISSUES = [
+    (
+        "<ISSUE DESCRIPTION PREFERABLY WITH GITHUB ISSUE REFERENCE>",
+        "<REGULAR EXPRESSION FOR GREP TO BE LOOKED FOR>",
+    ),
+    (
+        # The detection may be improved, not sure if just with grep.
+        "[1] https://github.com/rhinstaller/kickstart-tests/issues/675 The string is also in successful installation, but there is high chance that for test which times out it will be this issue.",
```
VladimirSlavik, line 30: I'm not sure I understand this - is it `[1]` or `[675]` ?

VladimirSlavik (approved):
Thank you, this should be very useful.

I am a bit confused by the message format, there's often multiple [numbers] or some in parens, but I guess that would be understandable once I actually looked at these issues?

scripts/classify-failures
rvykydal: I'll change it to [675] (the [1] is obsolete tracking)

rvykydal:
> Thank you, this should be very useful.
> 
> I am a bit confused by the message format, there's often multiple [numbers] or some in parens, but I guess that would be understandable once I actually looked at these issues?

Yes, it is free-form, ideally starting with [NNN] referring to the GH issue https://github.com/rhinstaller/kickstart-tests/issues/NNN. I put former tracking numbers (NNN < 100 in weekly reports) of the issue in parens. And multiple references are for issues that are perhaps the same issue in fact.
Any useful comment for a human looking at kstest results/logs can go to the message.

PR #1013, opened 2023-10-11
Fix & enable LUKS tests

jikortus:
The code copying RESULT file out of encrypted partitionings has been amended/fixed, so the LUKS tests work OK now.

The way guestfish is handled to do this task is somewhat cumbersome, but I couldn't find with a better solution. guestfish is not able to autodetect the partitioning and open/mount the LUKS container, so a series of explicit commands is needed to do so. I tried to combine this approach with --keys-from-stdin or --key <MD device>:key:passphrase, but neither option worked for me.

As I have only fairly limited experience with guestfish so far, I'm not able to tell whether my assumption on how to use the parameters is wrong, whether the options collide with 'manually' entering the commands, or this is a bug. If the latter is the case, I can create a BZ against libguestfs to cover this.

### DIFF
--- a/part-luks-1.sh
+++ b/part-luks-1.sh
@@ -17,11 +17,9 @@
 #
 # Red Hat Author(s): Vendula Poncova <vponcova@redhat.com>
 
-# On RHEL the test is manual because of broken reading of the results from image.
-# Check the results on the running VM.
 # Ignore unused variable parsed out by tooling scripts as test tags metadata
 # shellcheck disable=SC2034
-TESTTYPE="skip-on-rhel storage partition luks"
+TESTTYPE="storage partition luks"
 
 . ${KSTESTDIR}/functions.sh
 
--- a/functions.sh
+++ b/functions.sh
@@ -125,6 +125,33 @@ copy_file_encrypted() {
     run_with_timeout ${COPY_FROM_IMAGE_TIMEOUT} "guestfish --keys-from-stdin --ro ${disks} -i copy-out ${file} ${dir}"
 }
 
+copy_file_encrypted_raid() {
+    disks="$1"
+    file="$2"
+    dir="$3"
+
+    # we only assume 1 RAID device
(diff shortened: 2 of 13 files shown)


part-luks-1.sh
```diff
@@ -17,11 +17,9 @@
 #
 # Red Hat Author(s): Vendula Poncova <vponcova@redhat.com>
 
-# On RHEL the test is manual because of broken reading of the results from image.
-# Check the results on the running VM.
 # Ignore unused variable parsed out by tooling scripts as test tags metadata
 # shellcheck disable=SC2034
-TESTTYPE="skip-on-rhel storage partition luks"
```
rvykydal, line 24: part-luks-* are failing for me for rhel, (I've just run them locally on rhel9 and all failed with RESULT not found). That is why I enabled them only on Fedora. Are they stable for you? I wonder if it can be just my environment (even though the tests are run in container)?

rvykydal: Ah, I've just realized that my kstest runner container is 17 months (!) old. Let me try with something more recent :)

rvykydal: Indeed, it works with current container!

rvykydal (approved):
Looks good to me, thank you!

part-luks-1.sh
rvykydal: So let's enable the tests and see. If they show up unstable again, we can disable them. I won't do any harm.

jikortus:
@jstodola I'm not sure if such a message would be visible in the test output. However, I just noticed I forgot there two debug messages which I should likely get rid of.

jikortus:
OK, now after a discussion with @jstodola, I found out I originally misunderstood his comment, which is obvious to me now that I looked clearer where it was located. I fixed the related piece of code to output the MD devices found to `RESULT` file if more or less than 1 device is detected.

jstodola (approved):
It looks good to me, thanks!

PR #1014, opened 2023-10-12
daily scenario: switch to the current rhel devel releases

rvykydal:
It's the time again (the previous switch: https://github.com/rhinstaller/kickstart-tests/pull/937)


jstodola (approved):
Thanks!

VladimirSlavik (approved):

PR #1019, opened 2023-10-30
Disable bridge-no-bootopts-net test temporarily on rhel8 (gh1018)

rvykydal:
Disable bridge-no-bootopts-net test on rhel8 until gh1018 is fixed.


M4rtinK (approved):
Looks good to me, thanks! :)

PR #1037, opened 2023-12-07
Add test for bootupd/bootc ostree containers

VladimirSlavik:
WIP, needs some more checking...

### DIFF
--- a/fragments/shared/validation/success_on_first_boot.ks
+++ b/fragments/shared/validation/success_on_first_boot.ks
@@ -1,5 +1,5 @@
 # Report success if no errors have been reported on the first boot.
-# Additional tests can be implemented in /usr/libexec/kickstart-test.sh
+# Additional tests can be implemented in /etc/kickstart-test.sh
 # by the kickstart test that includes this fragment. This file is empty
 # by default.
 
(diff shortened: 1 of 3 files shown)


poncovka (commented):
Hi! I think the main information that want from this test is a confirmation that the final system is bootable. Can you reuse here some parts of the `reboot` test to verify that?

VladimirSlavik:
Still WIP, can't get the validation to work.

jikortus:
Here's a couple of findings related to earlier discussions with @VladimirSlavik which should likely be taken into account when finishing the test:

- The container image used in the test doesn't likely support BTRFS by default (no dracut module in initrd), so it will likely fail to mount root fs.
- `/boot` partition may need to have a `boot` label set, so that kernel and initrd files are properly found by GRUB.
- not sure if GRUB will find kernel and initrd files if `/boot` doesn't reside on a separate partition

jikortus:
Taking over the test development from Vladimir, will continue working on it after some higher priority tasks I've on my plate right now are done.

rvykydal:
FYI, there is some tests renaming / consolidation undergoing: https://github.com/rhinstaller/kickstart-tests/pull/1054

rvykydal:
Follow-up in https://github.com/rhinstaller/kickstart-tests/pull/1096

PR #1045, opened 2024-01-26
Fix and enable LVM cache tests

vojtechtrefny:
LVM started adding _cpool to the internal cache pool LVs so the checks needs to be adjusted to this change.


rvykydal (approved):
LGTM, thank you!

rvykydal:
I also tested that it works with rhel8 and rhel9 platforms.

PR #1048, opened 2024-02-01
Fix kickstart tests with deprecated timezone options

poncovka:
Anaconda removed support for the timezone --isUtc, --ntpservers and --nontp kickstart options. These options were fully deprecated in Fedora 40. See: https://github.com/rhinstaller/anaconda/pull/5437

**TODO:**

- [x] Test with Fedora.
- [x] Test with RHEL 9.
- [x] Test with RHEL 8.

### DIFF
--- a/container.ks.in
+++ b/container.ks.in
@@ -19,7 +19,8 @@ bootloader --disabled
 # Disable the NTP service.
 keyboard us
 lang en_US.UTF-8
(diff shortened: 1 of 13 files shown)


rvykydal (approved):
Looks good to me, thank you.
I also tested the PR on rhel8, rhel9, master, and daily-iso.

PR #1065, opened 2024-02-22
Fix and enable DDF RAID tests

vojtechtrefny:
The issue with `validate` is the only problem I found with this test, otherwise it seems to be working OK on both Fedora and RHEL.


rvykydal (approved):
Looks good to me, thank you!
(As a separate follow-up we may be able to share the validation in `functions.sh`)

PR #1067, opened 2024-02-28
Collect virt-install log when running a test.

rvykydal:
The log is useful when the test fails with "Problem starting virtual install".

The log will not be collected from remote runners. The remote runners functionality (TEST_REMOTES enviroment variable) is not used since a few years ago as the distribution of the tests to runners would be done on higher level now.

As can be perhaps seen from the PR, kickstart tests would deserve some logging consolidation, but it is an issue of bigger size.


jkonecny12 (approved):
Thanks!

PR #1070, opened 2024-02-28
Add RHEL 10 platform support

rvykydal:
For RHEL 10 adds:
- default repositories
- platform ks fragments

~~With the PR it should be possible to run kstests with RHEL 10 boot.iso locally updating the KSTEST_URL - it is still empty, so marking as DRAFT.~~

PRs for adding rhel10 scenario for daily runs (https://github.com/rhinstaller/kickstart-tests/pull/1071) and workflow for test on PR on rhel-10 anaconda branch will follow.

### DIFF
--- a/fragments/platform/rhel10/payload/ostreecontainer.ks
+++ b/fragments/platform/rhel10/payload/ostreecontainer.ks
@@ -1,0 +1,1 @@
+ostreecontainer --no-signature-verification --transport=registry --url=quay.io/centos-bootc/centos-bootc:stream10
--- a/fragments/platform/rhel10/validation/unified.ks
+++ b/fragments/platform/rhel10/validation/unified.ks
@@ -1,0 +1,6 @@
+# Validate installed packages on RHEL 10 system.
+
+# Check that the anaconda package was correctly installed.
+if ! rpm -q anaconda ; then
+    echo '*** anaconda package was not installed!' >> /root/RESULT
+fi
--- a/scripts/defaults-rhel10.sh
+++ b/scripts/defaults-rhel10.sh
@@ -1,0 +1,6 @@
+# Default settings for testing RHEL 10. This requires being inside the Red Hat VPN.
+
+source network-device-names.cfg
+export KSTEST_URL='http://download.eng.bos.redhat.com/rhel-10/nightly/RHEL-10-Public-Beta/latest-RHEL-10.0/compose/BaseOS/x86_64/os/'
+export KSTEST_MODULAR_URL='http://download.eng.bos.redhat.com/rhel-10/nightly/RHEL-10-Public-Beta/latest-RHEL-10.0/compose/AppStream/x86_64/os/'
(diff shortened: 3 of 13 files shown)


fragments/platform/rhel10/validation/unified.ks
```diff
@@ -0,0 +1,6 @@
+# Validate installed packages on RHEL 10 system.
```
KKoukiou, line 1: Are these copied from somewhere? Should they maybe be symlinks to some common files? 

Just asking for possible followup improvements, not nacking this :)

scripts/defaults-rhel10.sh
```diff
@@ -0,0 +1,6 @@
+# Default settings for testing RHEL 10. This requires being inside the Red Hat VPN.
+
+source network-device-names.cfg
+export KSTEST_URL='http://download.eng.bos.redhat.com/rhel-10/nightly/RHEL-10/latest-RHEL-10.0.0/compose/BaseOS/x86_64/os/'
+export KSTEST_MODULAR_URL='http://download.eng.bos.redhat.com/rhel-10/nightly/RHEL-10/latest-RHEL-10.0.0/compose/AppStream/x86_64/os/'
+export KSTEST_FTP_URL='ftp://ftp.tu-chemnitz.de/pub/linux/fedora/linux/development/rawhide/Everything/$basearch/os/'
```
KKoukiou, line 6: Is this non-standard mirror ok for the ftp?

KKoukiou (approved):

fragments/platform/rhel10/validation/unified.ks
rvykydal: Maybe, I think we haven't been using (modifying) this structure enough to start thinking about it yet (copied the stuff so far).
This fragment would be shared between three rhel platforms (rhel8, rhel9, rhel10 ) so maybe we could create a rhel folder as a target for symlinks (or link to rhel8?).
I'd be a bit afraid of trying to be generic at too high cost at this point, but we can track it as a potential follow-up issue.

jstodola (approved):
It looks good to me.

fragments/platform/rhel10/payload/ostreecontainer.ks
```diff
@@ -0,0 +1 @@
+ostreecontainer --no-signature-verification --transport=registry --url=quay.io/centos-bootc/centos-bootc:stream10
```
jkonecny12, line 1: This is not available and probably won't be anytime soon, so tests depending on this fragment will fail.

jkonecny12 (commented):
Are we fine with the broken link for ostree tests?

rvykydal:
> Are we fine with the broken link for ostree tests?

Yes we are, first lets's run the tests, then look at what is failing, why, and update or disable tests, or create issues, for rhel10

fragments/platform/rhel10/payload/ostreecontainer.ks
rvykydal: That's ok, we will see it and address it separately.

PR #1078, opened 2024-03-14
Ignore kernel deprecated driver warnings in the log monitor (gh#1077)

rvykydal:
We hit it first time with team driver, then after about one year with qla4xxx. It seems to be safe to ignore such messages.


jkonecny12:
I wonder if having an explicit list is not better. So we don't miss deprecation which can have impact on us in the future. 🤔

rvykydal:
> I wonder if having an explicit list is not better. So we don't miss deprecation which can have impact on us in the future. 🤔

I would be fine with explicit list, I am parted on this as well. The issue does not happen so often to be too big nuisance. Maybe let's make it generic next time?

jstodola (approved):
It looks good do me.

M4rtinK (approved):
Looks good to me as well. :)

rvykydal:
> > I wonder if having an explicit list is not better. So we don't miss deprecation which can have impact on us in the future. 🤔
> 
> I would be fine with explicit list, I am parted on this as well. The issue does not happen so often to be too big nuisance. Maybe let's make it generic (reconsider) next time?

I updated the PR to keep explicit list.

PR #1085, opened 2024-03-19
Update repo-install for added rhel10 scenario

rvykydal:


jstodola (approved):
It looks good to me.

jkonecny12 (approved):

scripts-lib.sh
```diff
@@ -6,7 +6,20 @@
 function is_rhel8 {
    local os_name=$1
    local os_version=$2
-   if [ ${os_name} == "rhel" -a ${os_version:0:1} == "8" ] ; then
+   if [ ${os_name} == "rhel" ] && [ ${os_version:0:1} == "8" ] ; then
+       echo "yes"
+   else
+       echo "no"
+   fi
```
jkonecny12, line 13: Wouldn't it be better to return `0` and have it as standard if instead of string comparation?

Or maybe even better to return the major version so you can have only one function and compare it to major version number.

rvykydal: I'll try to do the former as a follow-up.

rvykydal: https://github.com/rhinstaller/kickstart-tests/pull/1094

PR #1086, opened 2024-03-19
Disable some tests that on rhel10

rvykydal:


jstodola (approved):

jkonecny12 (approved):
Looks good to me. Although, it would be great to fix those failing tests.

PR #1094, opened 2024-03-21
Update is_rhelX function to return 0/1 values

rvykydal:


jkonecny12 (approved):
Looks good to me. I would probably prefer to have one function to return number of the major version and just do the compare test for the version but your solution is fine too 😉

PR #1100, opened 2024-03-22
workflows: run test-platforms status also when tests fail

rvykydal:
But still do not run it on any comment.

fix of commit 2d382a033660d448d7ad54f5c0c3c3eff62a1ab8

### DIFF
--- a/.github/workflows/test-platforms.yml
+++ b/.github/workflows/test-platforms.yml
@@ -327,7 +327,7 @@ jobs:
 
   result:
     needs: [platform, pr-info]
-    if: needs.pr-info.outputs.allowed_user == 'true'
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
Looks good to me, however, I'm thinking if merging the `result` job with the `platform` job wouldn't make your file easier. Is there a benefit of having these split?

rvykydal:
> Looks good to me, however, I'm thinking if merging the `result` job with the `platform` job wouldn't make your file easier. Is there a benefit of having these split?

Platform is a matrix job and result sets overall result of the platform matrix job (of all the jobs run in parallel) so seems natural to me as being the next job. I mean, how could platform report its own result?

PR #1113, opened 2024-03-25
Disable tests failing on newly added rhel10 scenario

rvykydal:
The issues are tracked in:

gh1104
gh1106
gh1105
gh1108
gh1109
gh1107
gh1110


rvykydal:
The test-platforms test should be OK, for some reason daily-iso download failed, but rawhide passed.

jkonecny12 (approved):

PR #1117, opened 2024-03-27
Enhance manual test execution

jikortus:
- pass KSTEST_* variables into the container env
- ensure disk images are kept with $KSTESTS_KEEP == 2

### DIFF
--- a/containers/runner/run-kstest
+++ b/containers/runner/run-kstest
@@ -93,8 +93,10 @@ ${KSTESTS_DIR}/scripts/log2json --scenario $SCENARIO --output $TEST_LOG.json $TE
 # Fixup permissions
 chmod -R a+rw /var/tmp/kstest-*
 # Clean up (artifacts from broken test)
-find /var/tmp/kstest-* -name "*.iso" -delete
-find /var/tmp/kstest-* -name "*.img" -delete
+if [ "$KSTESTS_KEEP" != "2" ]; then
+    find /var/tmp/kstest-* -name "*.iso" -delete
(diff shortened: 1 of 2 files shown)


rvykydal (approved):
Looks good to me, thank you.

containers/runner/run-kstest
```diff
@@ -93,8 +93,10 @@ ${KSTESTS_DIR}/scripts/log2json --scenario $SCENARIO --output $TEST_LOG.json $TE
 # Fixup permissions
 chmod -R a+rw /var/tmp/kstest-*
 # Clean up (artifacts from broken test)
-find /var/tmp/kstest-* -name "*.iso" -delete
-find /var/tmp/kstest-* -name "*.img" -delete
+if [ "$KSTESTS_KEEP" != "2" ]; then
+    find /var/tmp/kstest-* -name "*.iso" -delete
+    find /var/tmp/kstest-* -name "*.img" -delete
+fi
```
jkonecny12, line 99: Wouldn't this increase required size a lot for the artifacts? If yes, we shouldn't use this on GH Actions.

jkonecny12 (approved):
Looks good to me otherwise.

containers/runner/run-kstest
jikortus: IMO `$KSTESTS_KEEP` should default to 1 unless explicitly set to other value, so the disk images shouldn't be kept by default. I introduced this change because there was no way (unless I missed it) to keep the images, which may be crucial for debugging or writing new tests in some cases.

Issue #1120, opened 2024-04-03
daily-iso boot.iso uses "fedora" profile while rawhide boot.iso uses "fedora-server"


rvykydal:
While for our daily built iso (https://github.com/rhinstaller/kickstart-tests/actions/workflows/daily-boot-iso-rawhide.yml) the `fedora` profile is detected, for regular rawhide boot.iso test (https://github.com/rhinstaller/kickstart-tests/blob/17a749ac9b8887ad3936a895cf872c962e0dfe1d/containers/runner/launch#L115) `fedora-server` is the profile.

One consequence of it can be seen here: https://github.com/rhinstaller/kickstart-tests/pull/1116#issuecomment-2033780167

rvykydal:
Seems that using Everything boot.iso (https://mirror.karneval.cz/pub/linux/fedora/linux/development/rawhide/Everything/x86_64/os/images/boot.iso) would result in 'fedora' profile.

jstodola:
Nothing to do. Due to https://github.com/rhinstaller/kickstart-tests/pull/1123, both daily iso and rawhide use the `server` profile now. Closing after confirmation with @rvykydal .

PR #1123, opened 2024-04-03
Use Fedora Everything image instead of Fedora Server as the default i…

rvykydal:
…mage

Fedora Everyting is a generic 'fedora' variant from Anaconda profile configuration point of view, which we should test by default rather than Fedora Server which is a 'fedora-server' profile configuration inherited from 'fedora'. Specifically, 'fedora-server' is overriding default btrfs autopartitioning by lvm.

See https://github.com/rhinstaller/kickstart-tests/issues/1120


jkonecny12 (approved):
I agree. Let's try to be as much generic as possible.

PR #1124, opened 2024-04-04
Decrese the default timeout for a test to 30 minutes.

rvykydal:
Currently a test runs usually about 10 minutes or less. The infrastructure got pretty stable (the whole suite of 237 tests runs in 1h45m) so 30 minutes should be enough. This change will also decrease overall time to run the suite by not blocking the queue with timeouting tests (usually flakes) too long.

### DIFF
--- a/groups-and-envs-2.sh
+++ b/groups-and-envs-2.sh
@@ -24,7 +24,8 @@ TESTTYPE="packaging skip-on-rhel"
 
 . ${KSTESTDIR}/functions.sh
 
-# The test installs ~1366 packages for which 1h timeout is sometimes not enough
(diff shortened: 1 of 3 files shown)


groups-and-envs-2.sh
```diff
@@ -26,5 +26,5 @@ TESTTYPE="packaging skip-on-rhel"
 
 # The test installs ~1366 packages for which 1h timeout is sometimes not enough
```
jkonecny12, line 27: You should change this comment too.

jkonecny12 (approved):
Looks good to me otherwise.

PR #1130, opened 2024-04-09
Log timeouts of commands run in functions.sh

rvykydal:


KKoukiou (approved):
Are 10 seconds sufficient? Not sure what disks you copy there.

jkonecny12 (approved):

rvykydal:
> Are 10 seconds sufficient? Not sure what disks you copy there.

IIUIC the TERM signal timeout is given as parameter `${duration}` and only after sending this, KILL is sent after 10s.

PR #1131, opened 2024-04-09
Run virt-cat with timeout when probing boot iso

rvykydal:
Rather fail sooner than wait for timeout of the whole run batch which can be too long.


jkonecny12 (approved):
I wonder when is this happening.

PR #1137, opened 2024-04-16
Update kstest-runner container for Fedora 39 base image

rvykydal:


containers/runner/Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN dnf -y update && \
     rsync \
     git \
     virt-install \
+    libvirt-daemon \
```
KKoukiou, line 11: Not sure i follow but sure this is needed to run VMs.

rvykydal: It contains libvirtd, in F38 it was pulled in by some other virt package I guess, in F39 I think there was some packages consolidation in libvirt.

jkonecny12 (approved):
This is great! Thanks a lot!

PR #1138, opened 2024-04-19
Move logic from workflows to scripts

rvykydal:


scripts/generate-launch-args.py
```diff
@@ -0,0 +1,130 @@
...
+    _parser.add_argument("--skip-testtypes", "-s", type=str, metavar="TYPE[,TYPE..]",
+                         help="skip tests with TYPE (tag)")
+    _parser.add_argument("--testtype", "-t", type=str, metavar="TYPE",
+                         help="only run tests with TYPE (tag)")
+    _parser.add_argument("tests", nargs='*', metavar="TESTNAME",
+                         help="names of test to be run")
+    _parser.add_argument("--branch", "-b", type=str, metavar="GIT-BRANCH",
+                         help="Anaconda git branch (for example rhel-10)")
+    _parser.add_argument("--os-variant", "-r", type=str, metavar="OS_VARIANT",
+                         help="os variant of the boot.iso to be tested (for example rhel9)")
```
jkonecny12, line 85: Are these exclusive? If yes, it should be mentioned somewhere.

jkonecny12 (approved):
Looks good to me. I'm just not sure about the note below.

scripts/generate-launch-args.py
rvykydal: Good point, they are, fixed by making them exclusive for the parser.

PR #1141, opened 2024-04-24
Fix raid-1 test (gh#777)

rvykydal:
We want to test raid even when reqpart is not working for raid.  Will create issue and separate raid-1-requpart test using the reqpart command.


rvykydal:
The first commit just fixes the tests on rhel-10 and will be removed.

jkonecny12 (approved):

jstodola (approved):

PR #1150, opened 2024-05-13
Introduce gating with static code checks

KKoukiou:


jkonecny12:
Our experience is better with pylint. Ruff is great because it's fast but pylint is able to found more usually.

rvykydal (approved):
I'd be OK merging this and eventually open PR for using pylint that may or not replace this one.

PR #1153, opened 2024-05-13
Use symlinks for fragments folder

### LINKED ISSUE #1121: Use symlinks for fragments
Instead of copying content per platform folder (ie rhelX) we could use symbolic links.

For example in case when we need a modification for `fedora_rawhide` only, the `rhelX` variants can be symlinks to the default common fragments.

KKoukiou:
Fixes #1121


M4rtinK:
I'm afraid I don't really like this change.

Before, it was enough to check the base version for the fragment in the `shared` folder and then check if the given platform has an override for it. Now we are basically adding a third layer on top of that.

Also with the symlinks in place, I'm afraid one can also very easily edit the shared rhel fragment by mistake. Same thing when I would like to do an actual override for the RHEL version - one needs to first make sure to remove the symlink & create a new file in its place.

As we don't need to maintain many RHEL variant for kickstart tests anyway (basically just 9 & 10 going forward) I don't think the symlinks are necessary and the small amout of duplication is safer & simpler.

fragments/platform/rhel/repos/unified.ks
```diff
@@ -1,3 +1,3 @@
-# Unified ISO for RHEL 9 served on http server
+# Unified ISO for RHEL served on http server
 # BaseOS and AppStream repositories should be loaded automatically from the unified repo
 url RHEL9_UNIFIED_REPO_URL_GOES_HERE
```
rvykydal, line 3: RHEL9 -> RHEL ? (it is in the rhel)

rvykydal:
I feel a little bad because I myself created the issue https://github.com/rhinstaller/kickstart-tests/issues/1121 and now have some doubts as well, especially I am not sure if having rhel common ancestor makes sense given there will be max 3 rhel releases of which the most recent one can have quite a lot in common with fedora.
Feels also less transparent to me, but it might be just a habit.
I think we need more opinions.

KKoukiou:
@rvykydal @M4rtinK I dont have strong opinion - let's close this and focus on more priority items.

Issue #769, opened 2022-08-16
module-X tests failing due to missing nodejs:16 and swig:4


rvykydal:
[virt-install.log](https://github.com/rhinstaller/kickstart-tests/files/9349998/virt-install.log)
There seems to be only single module in the rawhide repository currently:
https://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/Modular/x86_64/os/

rvykydal:
PR to disable the tests: https://github.com/rhinstaller/kickstart-tests/pull/770

PR #1159, opened 2024-05-14
Skip modularity tests on fedora

KKoukiou:
These are already not running as they are disabled, and modularity is deprecated on Fedora-41. So let's just stop running these to clean up the disabled stack.

https://github.com/pykickstart/pykickstart/pull/491

Fixes: #769


rvykydal (changes requested):
We should remove also the tag from the list of disabled tests: https://github.com/rhinstaller/kickstart-tests/blob/da8c7b5b0f9157f51410819db2fa2ecb7dadfa90/containers/runner/skip-testtypes#L19

rvykydal (approved):

PR #1160, opened 2024-05-14
Since Fedora-36 authselect automatically selects the distro default profile

KKoukiou:


M4rtinK (approved):
Looks good to me, thanks! :)

Issue #1106, opened 2024-03-25
packages-and-groups-1 failing on rhel10: "org.fedoraproject.Anaconda.Modules.Payloads:missing packages: kacst*"


rvykydal:
```
00:36:21,047 WARNING org.fedoraproject.Anaconda.Modules.Payloads:INFO:anaconda.modules.payloads.payload.dnf.dnf_manager:Including specs: ['@server-product-environment', '@core', '@container-management', 'kacst*', 'kernel', 'e2fsprogs', 'lvm2', 'xfsprogs', 'grub2', 'grub2-tools', 'chrony', 'langpacks-en', '@platform-kvm']
00:36:21,047 WARNING org.fedoraproject.Anaconda.Modules.Payloads:INFO:anaconda.modules.payloads.payload.dnf.dnf_manager:Excluding specs: ['buildah', 'grub2-efi*', 'kmod-kvdo', 'tmux', 'vdo']
00:36:21,537 WARNING org.fedoraproject.Anaconda.Modules.Payloads:ERROR:dnf:No match for argument: kacst*
[...]

PR #1161, opened 2024-05-14
Replace kacst glob with the existing google-noto* packages glob

KKoukiou:
kacst is not packaged / gragged anymore in RHEL10. [1]

[1] https://bugs.documentfoundation.org/show_bug.cgi?id=152376

Fixes: #1106


rvykydal:
We should remove the issue tag from the list in https://github.com/rhinstaller/kickstart-tests/blob/master/containers/runner/skip-testtypes

rvykydal (approved):

PR #1162, opened 2024-05-14
Introduce pylint in static-code-checks workflow

KKoukiou:
Example run: https://github.com/KKoukiou/kickstart-tests/actions/runs/9082280741


jkonecny12 (approved):
Some of these fixes are not a nice bugs in the code, so thanks a lot for adding pylint.

Also, interesting solution for starting the tests. :)

PR #1166, opened 2024-05-15
Enable selinux-contexts on rhel9 (gh#970)

jstodola:
The problem we saw in the past has been resolved in kexec-tools:
https://gitlab.com/redhat/centos-stream/rpms/kexec-tools/-/commit/401619f484f2b2e943ca4ecc8fd90e02978a75c2


rvykydal (approved):
Thank you!

PR #1167, opened 2024-05-15
Update permissions of some logs to be easily removable (gh#1165)

rvykydal:
It will prevent some failures on copying logs when running launch script both with superuser and regular user in a single checkout without cleaning the data/logs folder.


M4rtinK (approved):
Looks good to me, thanks! :)

PR #1168, opened 2024-05-15
Replace sshpass with except (gh#1109)

marusak:
sshpass is not shipped in rhel10 thus use simple except script to interactively pass password into ssh.


rvykydal (approved):
Looks great to me, thank you.

Issue #902, opened 2023-03-13
daily-iso infra issue: "curl: (6) Could not resolve host: null"


rvykydal:
[daily-iso-download-failure.txt](https://github.com/rhinstaller/kickstart-tests/files/10955692/daily-iso-download-failure.txt)

PR #1170, opened 2024-05-15
Fix: daily-iso infra issue: "curl: (6) Could not resolve host: null"

KKoukiou:
The artifacts are paginated, and it can happen when there is high traffic on the repository that 'images' artifact are hidden in some page > 1. This makes the fetching of the 'images' artifact to fail.
Use the 'name' query parameter to limit the artifacts to the ones we are interested in.

Fixes: #902


rvykydal (approved):
Looks good to me, thank you, great job!

PR #1181, opened 2024-05-17
Fixup documentation for PR comment triggers

M4rtinK:
Looks like an (old ?) workflow name has been still reffered from the main README file. So fix that and also reformat the trigger documentation a bit.

### DIFF
--- a/.github/workflows/test-os-variants.yml
+++ b/.github/workflows/test-os-variants.yml
@@ -1,11 +1,20 @@
 # Run kickstart tests in a PR triggered by a "/test-os-variants" command from an organization member
 #
-# /test-os-variants - execute all tests affected by the PR or smoke tests if there is no such.
-# Or specify tests by combination of test names and/or options:
-# /test-os-variants --testtype TESTTYPE --skip-testtypes TYPE[,TYPE..] TEST1 TEST2
+# To run all tests affected by the PR (or just smoke tests if the PR changes no tests)
+# use this in a PR comment:
 #
-# For dry run use:
-# /test-os-variants-dry
+#  /test-os-variants
+#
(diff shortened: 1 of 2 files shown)


KKoukiou (approved):

.github/workflows/test-os-variants.yml
```diff
@@ -1,11 +1,20 @@
...
+#
+# Alternatively it is also possible to specify tests to run or test to be skipped:
+#
+#  /test-os-variants --testtype TESTTYPE --skip-testtypes TYPE[,TYPE..] TEST1 TEST2
+#
+# Lastly there is a dry-run mode, that just indicates which tests would be run
+# (useful to check test grouping and overall infra health):
+#
+#  /test-os-variants-dry
+#
```
rvykydal, line 16: The comment really deserves improvements, I'd suggest some modifications
- be honest that we just really test changed tests (not affected, ie no inference here)
- I think the "to specify tests to run or test to be skipped" doesn't help to clear up the options in the usage below, so rather let user to figure it out from the usage or add more examples (in a separate PR). (Or maybe replace with "to specify tests to run and/or test to be skipped")
- Make clear that the dry mode can be used for all the usage (including the specific)
Which brings me to this:
```
# To run all tests changed by the PR (or just smoke tests if the PR changes no tests)
# use this in a PR comment:
#
#  /test-os-variants
#
# Alternatively it is also possible to specify tests to run explicitly:
#
#  /test-os-variants --testtype TESTTYPE --skip-testtypes TYPE[,TYPE..] TEST1 TEST2
#
# It is possible to run in a dry-run mode, that just indicates which tests would be run
# (useful to check test grouping and overall infra health):
#
#  /test-os-variants-dry ...
#
```

rvykydal (approved):

PR #1182, opened 2024-05-17
Move all tests to tests/ subfolder

KKoukiou:
Rename existing 'test' folder containing linter script to 'tools' to avoid confusion.


rvykydal:
Just a hint for testing of the massive change, it can be useful at various stages to
- use dry run mode on some / all tests
- select just some tests for real test explicitly
(see the https://github.com/rhinstaller/kickstart-tests/blob/master/.github/workflows/test-os-variants.yml)

rvykydal:
One example of a location that will need to be updated I am aware of (I think tests would reveal others if there are any):
https://github.com/rhinstaller/kickstart-tests/blob/0def9ce503bf76db4ea5fa9b334ea306b16fd569/.github/workflows/scenarios-permian.yml#L46

KKoukiou:
Closing it, I don't plan to work on this now

Issue #1190, opened 2024-05-20
204/209 tests failing on rhel10: "org.fedoraproject.Anaconda.Modules.Payloads:missing packages: kdump-utils"


rvykydal:
https://issues.redhat.com/browse/RHEL-37138

rvykydal:
anaconda-conf logs:
[kstest.log](https://github.com/rhinstaller/kickstart-tests/files/15375092/kstest.log)
[virt-install.log](https://github.com/rhinstaller/kickstart-tests/files/15375094/virt-install.log)

rvykydal:
workaround for the issue, use "%packages --ignoremissing"

jstodola:
This issue is fixed in RHEL-10.0-20240520.72.

PR #1191, opened 2024-05-20
Switch rhel10 platform to centos images (gh#1190)

rvykydal:
Latest rhel10 nightly is broken
https://github.com/rhinstaller/kickstart-tests/issues/1190

### DIFF
--- a/scripts/defaults-rhel10.sh
+++ b/scripts/defaults-rhel10.sh
@@ -3,7 +3,7 @@
 # This is a reasonable default used until the new release is detected by osinfo library.
 export KSTEST_OSINFO_NAME=fedora-eln
 source ./network-device-names.cfg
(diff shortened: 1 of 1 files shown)


marusak:
> [radek wrote](https://github.com/rhinstaller/kickstart-tests/issues/1190#issuecomment-2120121082) workaround for the issue, use "%packages --ignoremissing"

Wouldn't that be better?

rvykydal:
> > [radek wrote](https://github.com/rhinstaller/kickstart-tests/issues/1190#issuecomment-2120121082) workaround for the issue, use "%packages --ignoremissing"
> 
> Wouldn't that be better?

Yes perhaps, I think it is for a separate PR and maybe some discussion, but indeed instead of a few ignoremissing tests we could have a few non-ignoremissing tests. We need to be a bit careful here with such a massive change but in general makes sense to me.

jstodola:
The problem with missing `kdump-utils` is fixed in the latest RHEL-10 nightly RHEL-10.0-20240520.72, is this change still needed?

rvykydal:
Thank you for heads-up, closing.

PR #1210, opened 2024-05-30
Add harddrive-iso-single test and "harddrive" tests updates

rvykydal:


rvykydal:
@jkonecny12 Could you please take a look at the PR ?

jkonecny12 (approved):
Looks good to me.

pre-lib-harddrive.sh
```diff
@@ -59,3 +55,22 @@ function prepare_tree() {
...
+    mkfs.ext4 -F "${disk}"
+}
+
+
+# Format single partition on a disk for hdd payload source
+function format_single_partition () {
+    local disk="$1"
+
+    sgdisk --zap-all ${disk}
+    sgdisk --new=0:0:+12GiB ${disk}
```
jstodola, line 74: Shouldn't the partition size be also increased to 22GiB?

rvykydal: Good point, let me check, we should probably parametrize the function.

rvykydal: Oh, I guess we need the unallocated space for autopart.

rvykydal:
I removed disabling on fedora as the upstream patch https://github.com/rhinstaller/anaconda/pull/5687 is already merged (daily-iso should pass, rawhide still fail)

rvykydal:
> I am removing disabling on fedora as the upstream patch [rhinstaller/anaconda#5687](https://github.com/rhinstaller/anaconda/pull/5687) is already merged (daily-iso should pass, rawhide still fail)

Indeed: https://github.com/rhinstaller/kickstart-tests/actions/runs/9695026372

jstodola (approved):
It looks good to me.

PR #1225, opened 2024-06-19
Switch hostname for downloading RHEL

jstodola:
The bos lab will be decommissioned soon. Switch to the devel/rdu lab.


rvykydal (approved):
Looks good to me, thank you.

PR #1228, opened 2024-06-24
Update packages instLangs and excludeWeakdeps options (gh#1227)

rvykydal:
In https://github.com/pykickstart/pykickstart/pull/494 the options were removed on rhel10.


rvykydal:
Note: I was also thinking about just skipping the test on rhel8 (skip-on-rhel-8) as its development ends, but I prefer to clean this (rhel8  special cases) up globally when we are sure we want to do it.

jstodola (approved):

PR #1229, opened 2024-06-25
Get list of installer image packages when probing boot.iso

rvykydal:
You can see the change in action in the /test-os-variants workflow of this PR, step "Run kickstart tests in container" (near "Saving list of installer image packages to /var/tmp/kstest-image-packages.log")
or in logs / artefacts of the workflow (eg logs-daily-iso) where you can find kstest-image-packages.log.

### DIFF
--- a/scripts/run_kickstart_tests.sh
+++ b/scripts/run_kickstart_tests.sh
@@ -182,7 +182,10 @@ ISO_OS_NAME=$(echo "${output}" | grep 'NAME=')
 ISO_OS_NAME="${ISO_OS_NAME##NAME=}"
 ISO_OS_VERSION=$(echo "${output}" | grep 'VERSION=')
 ISO_OS_VERSION="${ISO_OS_VERSION##VERSION=}"
-
+ISO_PACKAGES=$(echo "${output}" | grep 'PACKAGES=')
+ISO_PACKAGES="${ISO_PACKAGES##PACKAGES=}"
+echo Saving list of installer image packages to /var/tmp/kstest-image-packages.log
+echo $ISO_PACKAGES | tr ' ' '\n' | tee /var/tmp/kstest-image-packages.log
 
 # Append sed args to substitute
 sed_args=" -e s#@KSTEST_OS_NAME@#${ISO_OS_NAME}# -e s#@KSTEST_OS_VERSION@#${ISO_OS_VERSION}#"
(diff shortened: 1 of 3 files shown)


KKoukiou (approved):
Very useful thanks.

scripts/run_kickstart_tests.sh
```diff
@@ -182,7 +182,10 @@ ISO_OS_NAME=$(echo "${output}" | grep 'NAME=')
 ISO_OS_NAME="${ISO_OS_NAME##NAME=}"
 ISO_OS_VERSION=$(echo "${output}" | grep 'VERSION=')
 ISO_OS_VERSION="${ISO_OS_VERSION##VERSION=}"
-
+ISO_PACKAGES=$(echo "${output}" | grep 'PACKAGES=')
+ISO_PACKAGES="${ISO_PACKAGES##PACKAGES=}"
+echo Saving list of installer image packages to /var/tmp/kstest-image-packages.log
+echo $ISO_PACKAGES | tr ' ' '\n' | tee /var/tmp/kstest-image-packages.log
```
jkonecny12, line 188: If I read this correctly then the list of packages will get printed to stdout. If this is correct I think it would spam the test results a lot wouldn't it? Isn't it enough to just store the list?

jkonecny12 (approved):
Looks good to me. Just a question below.

jkonecny12:
Yes, it seems to me that it creates a mess in GH actions console output:
https://github.com/rhinstaller/kickstart-tests/actions/runs/9663125740/job/26654555035#step:15:107

I think we should remove use of `tee` to keep the log clarity.

scripts/run_kickstart_tests.sh
jstodola: If the tests pass, you do not need to worry about extra lines in the output, if there is a failure it might be handy to see the package versions right in the test output. I actually like it this way :-)

jstodola (approved):

scripts/run_kickstart_tests.sh
rvykydal: I think in this case it is handy to have this in the log output without need of making sure we really pass / gather all the required logs (for example in GH actions). Also it is maybe more discoverable this way.

rvykydal:
> Yes, it seems to me that it creates a mess in GH actions console output: https://github.com/rhinstaller/kickstart-tests/actions/runs/9663125740/job/26654555035#step:15:107
> 
> I think we should remove use of `tee` to keep the log clarity.

I already replied earlier, also given we even already have the debugging shell output above I don't see it as a problem.

PR #1231, opened 2024-06-26
Allow unknown args to be passed to generate-launch-args

rvykydal:
So that we can use single line in workflow comment api, like in https://github.com/rhinstaller/anaconda/pull/5722


jkonecny12 (approved):
Looks good to me.

PR #1239, opened 2024-07-08
Remove dracut-visible-arnings check

r0x0d:
We removed the deprecated warnings for inst.* in anaconda, thus, this test will always assert for something that will not appear anymore

Anaconda PR: https://github.com/rhinstaller/anaconda/pull/5723


KKoukiou (approved):
@M4rtinK can you please take a look as well?

jkonecny12 (approved):
Looks good to me. Thanks!

rvykydal (commented):
@jkonecny12 shouldn't we rather modify the test so that is still tests (IIUIC it was the purpose of https://github.com/rhinstaller/kickstart-tests/commit/0f30bea02182d9b29ef350ec442092a3f5d77939) the feature: https://github.com/rhinstaller/anaconda/pull/3533 ?

jkonecny12:
@rvykydal if you mean to modify the test to test if the warnings are not showed then I think this is better solution. Let's spare resources of our kickstart tests here. If the warning was missed somewhere than it is not a critical issue but we would leave this test running for-ever.

Do you have some specific reason why would you leave it here what I might be missing?

rvykydal (approved):

rvykydal:
I just wanted to have a (generic) test for the feature used for the warnings but yes, let's handle it out of scope of this PR, if at all.

PR #1249, opened 2024-07-16
Use bigger disk for fedora-live-image-build test (gh1248)

rvykydal:
The test started to fail with "Not enough space in file systems for the current software selection. An additional 700 MiB is needed."

### DIFF
--- a/fedora-live-image-build.sh
+++ b/fedora-live-image-build.sh
@@ -23,6 +23,13 @@ TESTTYPE="packaging skip-on-rhel"
 
 . ${KSTESTDIR}/functions.sh
 
+prepare_disks() {
+    tmpdir=$1
+
+    qemu-img create -q -f qcow2 ${tmpdir}/disk-a.img 20G
+    echo ${tmpdir}/disk-a.img
+}
+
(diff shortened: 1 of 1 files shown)


fedora-live-image-build.sh
```diff
@@ -23,6 +23,13 @@ TESTTYPE="packaging skip-on-rhel"
 
 . ${KSTESTDIR}/functions.sh
 
+prepare_disks() {
+    tmpdir=$1
+
+    qemu-img create -q -f qcow2 ${tmpdir}/disk-a.img 12G
```
jstodola, line 29: AFAIK, `qemu-img` doesn't allocate the specified image size on the disk when creating it. Because of that I think the disk size should be defined even bigger (20G?), just to prevent having to increase the value in the near future again. I do not expect the size of the image will stop increasing :-)

jkonecny12: I agree, let's take a benefit of the qcow2 format.

rvykydal: Updated

jkonecny12 (approved):
Looks good to me.

PR #1252, opened 2024-07-16
Enable rpm-ostree-container tests on rhel-10

jkonecny12:
There is no reason to have these disabled on rhel-10. I guess it's here for historical reasons.


jkonecny12:
Rebased to resolve conflicts.

rvykydal (approved):

PR #1260, opened 2024-07-19
Add test support of bootupd EUFI entry insertion

jkonecny12:
Enhance existing test `rpm-ostree-container-bootc` to check for UEFI entry insertion on the first disk.
Add a new test working similar to other container tests but with `bootloader --leavebootorder` which will disable UEFI entry insertion.

To be able to do this add support to run VM in UEFI mode. And fix one bug during the path which caused troubles here.

The bootc test will be enabled only for Fedora until https://github.com/rhinstaller/anaconda/pull/5760 and https://github.com/rhinstaller/anaconda/pull/5761 are merged. The leavebootorder test should work all the time because UEFI entry just won't be inserted anyway on RHEL 9 and 10

### DIFF
--- a/functions.sh
+++ b/functions.sh
@@ -39,6 +39,13 @@ inject_ks_to_initrd() {
     echo "true"
 }
 
+enable_uefi() {
+    # Run the VM in the UEFI mode.
+    # This will add '--boot uefi' argument to the virt-install script
+    # returns: "true" or "false"
+    echo "false"
+}
+
 EXTRA_BOOTOPTS=$(echo "${KSTEST_EXTRA_BOOTOPTS}" | tr ';' ' ')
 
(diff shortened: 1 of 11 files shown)


rvykydal (approved):
Looks good to me, thank you.

jkonecny12:
**UPDATED**:
* Add new test *-uefi which is a copy of bootc test with the UEFI mode
* Remove bootc changes a leave the test as it was

The bootc test won't work in UEFI mode for some reason and honestly maybe that is expected because UEFI entry is not created by bootupd. So to be able to cover all the cases let's have another test for uefi which will be used only on Fedora where this functionality works.

rvykydal (approved):
Looks good to me.

PR #1265, opened 2024-07-23
Add support for platform specific ISOs

jkonecny12:
Currently, kickstart tests require to have ISO stored in `data/images/boot.iso` which is limiting and problematic.

The main issues with this approach:
- Switching between different platforms (`rhel9`, `fedora`...) means always changing the boot.iso manually.
- Harder to run multiple tests for different platforms in parallel because they all share the same `data/images/boot.iso`.
- You have to manually store ISO files when switching platforms or face a lot of downloading.
- You can't tell platform of the `data/images/boot.iso` file. Or at least it is harder to say just from looking on the file.

One solution to this issue is to have different `$PWD`, so the data directory will be different for each platform. However, then you have to make sure you are in a correct directory before start of the test. Also you will have manage multiple directories for log files.

This change is proposing better solution. When ISO file is downloaded it is downloaded to `data/images/boot-<platform>.iso` so it is specific for each platform. This resolves the issues above cleanly.

Also if `data/images/boot.iso` already exists it will use that so this change is backward compatible.

### DIFF
--- a/containers/runner/launch
+++ b/containers/runner/launch
@@ -1,7 +1,8 @@
 #!/bin/bash
 # Run a set of tests (positional command line arguments) in the runner container in podman or docker.
 # Tests are taken from the current kickstart-tests checkout
-# If data/images/boot.iso exists, that is tested, otherwise it downloads the current Fedora Rawhide version.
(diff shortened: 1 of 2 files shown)


rvykydal (approved):
Looks good to me, nice, thank you!

PR #1279, opened 2024-08-12
Fix RESULT validation

jkonecny12:
Currently if RESULT file has an issue after the "SUCCESS" word it will tell success. That works in most of the cases but not if validation is done in the %post of KS but also later in the .sh file.

If %post section was fine and add SUCCESS but validation section will found issue it would take the whole test as successful.

Fix this in all the impacted tests and also in the functions.sh.


M4rtinK (approved):
Looks good to me. :)

Issue #1298, opened 2024-09-04
rpm-ostree-container-uefi failing on daily-iso with "EFI boot entry wasn't created properly"


rvykydal:
Since 2024-09-03
Last success: https://github.com/rhinstaller/kickstart-tests/actions/runs/10658159832
First failure: https://github.com/rhinstaller/kickstart-tests/actions/runs/10674027365
(packages diff below)
```
2024-09-03 00:13:08,836 DEBUG: Failed to run subprocess: '['/opt/kstest/kickstart-tests/scripts/launcher/lib/launcher_interface.sh', '-i', '/opt/kstest/data/images/boot.iso', '-k', '1', '-w', '/var/tmp/kstest-rpm-ostree-container-uefi.2024_09_03-00_08_41.c6vsqdjn', '-t', '/opt/kstest/kickstart-tests/rpm-ostree-container-uefi.sh', 'validate']'
stdout:
EFI boot entry wasn't created properly:
Timeout: 0 seconds
BootOrder: 0001,0000
Boot0000* UiApp|FvVol(7cb8bdc9-f8eb-4f34-aaea-3ee4af6516a1)/FvFile(462caa21-7614-4503-836e-8ab6f4662331)
Boot0001* Fedora
|_______HD(1,GPT,8ac33c10-b314-4d9a-8859-28d723ef3170,0x800,0x12c000)/^[FI^Ledora\shimx64.efi
*** System has started with errors:
EFI boot entry wasn't created properly:
Timeout: 0 seconds
BootOrder: 0001,0000
Boot0000* UiApp|FvVol(7cb8bdc9-f8eb-4f34-aaea-3ee4af6516a1)/FvFile(462caa21-7614-4503-836e-8ab6f4662331)
Boot0001* Fedora
|_______HD(1,GPT,8ac33c10-b314-4d9a-8859-28d723ef3170,0x800,0x12c000)/^[FI^Ledora\shimx64.efi

2024-09-03 00:13:08,836 INFO: RESULT:rpm-ostree-container-uefi:b722a5642e59:FAILED:Validation failed with return code 1
```

[kstest.log](https://github.com/user-attachments/files/16861702/kstest.log)
[virt-install.log](https://github.com/user-attachments/files/16861703/virt-install.log)

rvykydal:
Packages diff: 
[lorax-packages.log.diff.txt](https://github.com/user-attachments/files/16861723/lorax-packages.log.diff.txt)
```
[...]

PR #1299, opened 2024-09-04
Disable rpm-ostree-container-uefi until gh1298 is fixed

rvykydal:

PR #1301, opened 2024-09-04
Fix validation issue in rpm-ostree-container-uefi

jkonecny12:
The current efibootmgr validation has failed because newer uefi entries are so long that the line for the given entry is split to two lines.

Fix this by removing newlines and tweaking the grep calls.

Closes https://github.com/rhinstaller/kickstart-tests/issues/1298
Obsoletes https://github.com/rhinstaller/kickstart-tests/pull/1299


rvykydal (approved):
Looks good to me, thank you.

PR #1302, opened 2024-09-13
Tests for kickstart script commands

adamkankovsky:
Tests for PR: https://github.com/rhinstaller/anaconda/pull/5739

### DIFF
--- a/script-post.ks.in
+++ b/script-post.ks.in
@@ -1,0 +1,32 @@
+# Test name: post_install_with_logging_test
+%ksappend repos/default.ks
+%ksappend common/common_no_payload.ks
+%ksappend payload/default_packages.ks
+
+# System configuration for automatic installation
+text
+
+# Post-install script (nochroot)
+%post --nochroot --log=/mnt/sysroot/root/post-nochroot.log
+# Log to the nochroot log file
+echo "Post-install (nochroot) script started"
+# Write 'SUCCESS' to the RESULT file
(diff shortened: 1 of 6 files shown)


rvykydal (commented):
This is great work, thank you!
I have some suggestions regarding naming, metadata and reusing:

- I'd rather make all the tests of common `TESTTYPE`, something like `ksscript` or `script` so that it is easy to run them together (like in the anaconda PR: `/kickstart-test --kstest-pr 1302 --testtype ksscript`)
- I'd name the tests with common prefix, like `script-pre` instead of `pre-script`
- I think we can use fragments (`%ksinclude`) more, like in [ui_text_interactive](https://github.com/rhinstaller/kickstart-tests/blob/ed5cddaeaad042c1c9ff1e7d4368cd38866bb679/ui_text_interactive.ks.in#L4). This can be done as a follow-up though.

rvykydal (commented):
The update looks good to me, however
- one test is failing on testing farm, not sure why
- the tests are not run in permian (on /test-os-variants)) because of the .sh files not being executable (https://github.com/rhinstaller/kickstart-tests/blob/85f08565ae37e979f071e07ecceb55fa29528c05/scripts/generate-testcases.py#L72, example of a test making executable: https://github.com/rhinstaller/kickstart-tests/commit/51d6ecbf309543848eb51584a6282bc0604df5b1). Sorry for this non-transparent requirement.

rvykydal (approved):
Looks good to me, great job, thank you.

PR #1308, opened 2024-09-20
Extend payload group

jkonecny12:
Extend the `payload` group to all payload related tests and remove old `method` group which is outdated and confusing.


M4rtinK (approved):
Looks good to me. :)

rvykydal (approved):
Looks good to me, thank you!

jkonecny12:
Tests are failing but this PR is not modifying tests directly so it seems to be something different. Also test-os-variants is running even known-issue and similar tests.

PR #1315, opened 2024-10-09
Update classify script for gh1311 gh1312 gh1313 gh1314

rvykydal:


scripts/classify-failures
```diff
@@ -186,15 +178,53 @@ ISSUES = [
...
+        "description": "[1314] https://github.com/rhinstaller/kickstart-tests/issues/1314",
+        "first_grep": "CRIT kernel:virtio_net virtio1 enp1s0: NETDEV WATCHDOG: CPU: 0: transmit queue 0 timed out",
+    },
+    {
+        "description": "[1318] https://github.com/rhinstaller/kickstart-tests/issues/1318",
+        "first_grep": "anaconda:anaconda: display: X or window manager startup failed: systemd exited with status 1",
+    },
+    {
+        "description": "[1319] https://github.com/rhinstaller/kickstart-tests/issues/1319",
+        "first_grep": "Anaconda.Modules.Payloads:ERROR:dnf.rpm:Error in POSTIN scriptlet in rpm package",
```
jstodola, line 213: I suggest to have the whole error message here to make sure it matches only the error caused by the `ca-certificates` package. To make sure a problem in a different package is not overlooked.

rvykydal: It would mean to split out the problem for selinux-policy-targeted package from the https://github.com/rhinstaller/kickstart-tests/issues/1319 to a separate issue, right?

jstodola: Yes, I think they should should be separate issues - they would be most likely fixed in different components.

rvykydal: So I rather removed [1319] from this PR.
I will watch the results some time and decide then if https://github.com/rhinstaller/kickstart-tests/issues/1319 should be really split.

jstodola (approved):
It looks good to me.

PR #1321, opened 2024-10-15
Add an UEFI test (reboot-uefi) to coverage and smoke groups

rvykydal:
Resolves: INSTALLER-4061
A modification of `reboot` test.

### DIFF
--- a/reboot-uefi.ks.in
+++ b/reboot-uefi.ks.in
@@ -1,0 +1,25 @@
+#version=DEVEL
+#test name: reboot-uefi
+
+# Use defaults.
+%ksappend repos/default.ks
+%ksappend common/common_no_payload.ks
+%ksappend payload/default_packages.ks
+
+# Reboot the installed system.
+reboot
+
+# Validate on the first boot.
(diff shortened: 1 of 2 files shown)


rvykydal:
/test-os-variants --testtype smoke --skip-testtypes skip-on-rhel-10

jstodola (approved):
It looks fine.
If I should suggest an enhancement, it would be nice to check that the system is really running in UEFI mode - either during the installation or after reboot. Perhaps just a simple execution of `efibootmgr` should be enough (it fails on non-UEFI systems).

rvykydal:
I should mention that this is a modification of `reboot` test.

rvykydal:
> It looks fine. If I should suggest an enhancement, it would be nice to check that the system is really running in UEFI mode - either during the installation or after reboot. Perhaps just a simple execution of `efibootmgr` should be enough (it fails on non-UEFI systems).

@jstodola updated

jstodola (approved):
It looks good, thanks.

PR #1323, opened 2024-10-21
Extend container-uefi with fix for vconsole keymap

jkonecny12:
It is fix for
https://bugzilla.redhat.com/show_bug.cgi?id=1890085 https://issues.redhat.com/browse/INSTALLER-2537

Currently, virtual console is not correctly set from localed on an ostree installations. To resolve this issue let's put the `vconsole.keymap` to the installed system kernel arguments.

Depends on  https://github.com/rhinstaller/anaconda/pull/5952

### DIFF
--- a/rpm-ostree-container-uefi.ks.in
+++ b/rpm-ostree-container-uefi.ks.in
@@ -59,9 +59,15 @@ if [ ${?} -ne 0 ]; then
 fi
 
 if [ "${remote_url}" != "${expected_url}" ]; then
-    echo "Unexpected URL: ${remote_url}, expected ${expected_url}"" >> /root/RESULT
+    echo "Unexpected URL: ${remote_url}, expected ${expected_url}" >> /root/RESULT
 fi
 
+# Test for vconsole kernel argument (specific to ostree installations)
+# https://bugzilla.redhat.com/show_bug.cgi?id=1890085
+journalctl | grep -q "vconsole.keymap=us"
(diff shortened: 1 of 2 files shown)


jkonecny12:
Expected to fail without the PR but I want to verify that behavior.

jkonecny12:
The test-os-variants won't work because my patch was just merged so it is not yet part of the Rawhide. Do we want to wait for the release?

rvykydal:
> The test-os-variants won't work because my patch was just merged so it is not yet part of the Rawhide. Do we want to wait for the release?

Maybe let's wait for daily-iso and merge?

jkonecny12:
I see even the TMT is using Rawhide ISO. In that case we should probably wait for release. That is a pain honestly.

rvykydal:
> I see even the TMT is using Rawhide ISO. In that case we should probably wait for release. That is a pain honestly.

In general I'd be OK for merging just with
- the test passing on the Anaconda PR (/kickstart-test --kstest-pr 1323 rpm-ostree-container-uefi) to see it is actually working, and
- the test passing `/test/os/variants` to the extent we don't break daily tests (that means passing every variant we run in dailies - so `rawhide` not required. Waiting for `daily-iso` should be usually 1 day after merge so it seems fine to me.

jkonecny12:
Not blocked anymore. Changes are part of Rawhide now. @rvykydal could you please do a review of this PR?

PR #1325, opened 2024-10-29
Adjust luks-lvm-2 test case to cover RHEL-45180

vojtechtrefny:
We have a bug in blivet triggered by using LUKS encrypted LV with --size=1 and --grow. This can be easily covered by adjusting the existing luks-lvm-2 test case which also uses --grow.


vojtechtrefny:
This needs https://github.com/storaged-project/blivet/pull/1309 for the test to pass, marking as blocked for now.

jstodola (approved):

rvykydal (approved):

PR #1326, opened 2024-10-31
Stop travis runs by removing config file and update documentation

KKoukiou:
We run our tests reliably on testing farm and we stopped our travis subscription.

### DIFF
--- a/plans/testing-farm.fmf
+++ b/plans/testing-farm.fmf
@@ -37,7 +37,7 @@ discover:
         trap "cp -rv data/logs/ ${TMT_TEST_DATA}/" EXIT INT QUIT PIPE
 
         cd /root/kickstart-tests
-        scripts/run_travis.sh
+        scripts/run-ci.sh
 
 
       duration: 6h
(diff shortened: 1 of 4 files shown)


M4rtinK (approved):
Looks good to me. :)

plans/testing-farm.fmf
```diff
@@ -37,7 +37,7 @@ discover:
         trap "cp -rv data/logs/ ${TMT_TEST_DATA}/" EXIT INT QUIT PIPE
 
         cd /root/kickstart-tests
-        scripts/run_travis.sh
+        scripts/run.sh
```
jkonecny12, line 40: We should use a more specific name. I'm afraid of people running this to run kickstart-tests.
Maybe `run-in-tmt` or `run-ci`?

jkonecny12 (changes requested):
Looks good to me otherwise.

plans/testing-farm.fmf
KKoukiou: Good point.

PR #1330, opened 2024-11-06
Update pre-install test for new modularization

adamkankovsky:
Due to modularization it was necessary to make a change in the tests because the triggered error from the module shows a traceback in the log, which automatically terminates the test at --erroronfail. So the temporary solution is to test the test without this parameter, and look for the error in the log.


rvykydal (approved):
Looks good to me. Agree with the plan.

PR #1331, opened 2024-11-07
Ignored simple tests

adamkankovsky:


scripts/launcher/lib/log_monitor/log_handler.py
```diff
@@ -42,6 +44,16 @@ class VirtualLogRequestHandler(LogRequestHandler):
...
 
+    # Path to the config file
+    CONFIG_FILE_PATH = '/tmp/ignored_simple_tests.conf'
+
+    # Read additional ignored messages from the config file
+    if os.path.isfile(CONFIG_FILE_PATH):
+        with open(CONFIG_FILE_PATH, 'r') as config_file:
+            extra_ignored_tests = [line.strip() for line in config_file if line.strip()]
+            # Extend the ignored_simple_tests list
+            ignored_simple_tests.extend(extra_ignored_tests)
```
jkonecny12, line 55: This solution is not inline with what we are doing elsewhere and also it will be taken by all the tests in parallel because the file is available to all.

adamkankovsky: You're right, it didn't occur to me that it actually behaves that way. I've edited the code and it should be fine now. I've tested everything.

KKoukiou (commented):
Can you revert the merged PR with the adjust pre-install test as part of this PR so that we can test that it works :)

PR #1341, opened 2024-11-22
Remove #test name comment from ks.in files

rvykydal:
I am not aware of any purpose of this comment and it is difficult to keep in sync with the real test name. If at all it should be generated at some point.

For example in https://github.com/rhinstaller/kickstart-tests/pull/1324 I suggested renaming of the test in the review, but the `#test name` was not updated during update of the PR. Instead requiring additional update I decided to handle it in an additional PR or to get rid of the comments completely.

### DIFF
--- a/anabot-1.ks.in
+++ b/anabot-1.ks.in
@@ -1,4 +1,3 @@
 #version=DEVEL
-#test name: anabot-1
 
 # NOTHING TO DO HERE! THIS FILE IS NOT USED!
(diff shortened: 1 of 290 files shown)


KKoukiou (approved):

M4rtinK:
I think I faintly remember this was used to check what test you are actually running by looking at the kickstart the given VM is using. Otherwise in absence of other metadata, you only see the kickstart contents. But I'm not sure if this is really that relevant anymore with the way we run kickstart tests & even name the test VMs based on the test being run.

If we really need to, we can always quite easily generate the test name (and possibly other stuff) when we compile the test kickstart from fragments.

Issue #1349, opened 2025-01-10
network static tests failing due to NM default profile update with "*** Failed check: IPADDR=10.0.2.200 ..."


rvykydal:
First hit 2025 Jan 10
Caused by https://github.com/NetworkManager/NetworkManager/commit/38dca2f044a5cf0f68752d3d73ceb5407f948238
```
2025-01-10 01:29:37,885 DEBUG: Failed to run subprocess: '['/opt/kstest/kickstart-tests/scripts/launcher/lib/launcher_interface.sh', '-i', '/opt/kstest/data/images/boot.iso', '-k', '1', '-w', '/var/tmp/kstest-network-static-httpks.2025_01_10-01_20_45.18tq2shr', '-t', '/opt/kstest/kickstart-tests/network-static-httpks.sh', 'validate']'
stdout:
*** Failed check: IPADDR=10.0.2.200 in /etc/sysconfig/network-scripts/ifcfg-enp2s0 or ipv4.address1=10.0.2.200/24,10.0.2.2 in /etc/NetworkManager/system-connections/enp2s0.nmconnection

2025-01-10 01:29:37,886 INFO: RESULT:network-static-httpks:d369755519f5:FAILED:Validation failed with return code 1
```
[kstest.log](https://github.com/user-attachments/files/18372797/kstest.log)
[virt-install.log](https://github.com/user-attachments/files/18372803/virt-install.log)

rvykydal:
PR to update only rhel-10:  https://github.com/rhinstaller/kickstart-tests/pull/1354

PR #1350, opened 2025-01-10
Update network static tests for NM change (gh1349)

rvykydal:
NM now dumps ipv4 address setting value into two separate 'address1' and 'gateway' keys by default.
https://github.com/NetworkManager/NetworkManager/commit/38dca2f044a5cf0f68752d3d73ceb5407f948238

Fixes: https://github.com/rhinstaller/kickstart-tests/issues/1349


M4rtinK (approved):
Looks good to me. :)

PR #1351, opened 2025-01-10
Check systemd-resolved package presence in dns test (rhbz#2304080)

rvykydal:
This should warn us about systemd-resolved support inconsistencies in Anaconda and target system.


M4rtinK (approved):
Looks good to me. :)

PR #1373, opened 2025-02-25
Add second test for LVM on RAID

vojtechtrefny:
This and the change to the first test case make sure the MD array can be referred by both its name and the "mount point" when used under a VG and also make sure the preferred name is not changed by or pykickstart.

This needs new version of both pykickstart and anaconda. See https://github.com/rhinstaller/anaconda/pull/6185 and https://github.com/pykickstart/pykickstart/pull/534

### DIFF
--- a/containers/runner/skip-testtypes
+++ b/containers/runner/skip-testtypes
@@ -57,6 +57,7 @@ rhel9_skip_array=(
   gh641       # packages-multilib failing on systemd conflict
   gh774       # autopart-luks-1 failing
   gh804       # tests requiring dvd iso failing
+  rhel75659   # pykickstart removes md_ prefix from new MD devices
 )
 
 rhel10_skip_array=(
@@ -70,6 +71,7 @@ rhel10_skip_array=(
   gh1110      # storage-multipath-autopart failing
   gh1207      # packages-multilib failing
   gh1213      # harddrive-iso-single failing
+  rhel80086   # pykickstart removes md_ prefix from new MD devices
 )
 
 # used in workflows/daily-boot-iso-rhel8.yml
--- a/lvm-raid-1.sh
+++ b/lvm-raid-1.sh
@@ -19,7 +19,7 @@
 
 # Ignore unused variable parsed out by tooling scripts as test tags metadata
 # shellcheck disable=SC2034
-TESTTYPE="lvm storage gh1414"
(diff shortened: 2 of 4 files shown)


vojtechtrefny:
Marking as ready for review, new versions of both pykickstart and anaconda are now released and either already available in rawhide (pykickstart) or on its way to rawhide (anaconda).

rvykydal (commented):
A typo on the commit message:
"... and also make sure the preferred name is not changed
by or pykickstart."

How should we deal with failures on rhel-9 and rhel-10 (/test-os-variants)?
- disable them with skip-on-rhel
- disable them with a reference to an jira or github issue and update the disabled tests table?, and reenable when they hit rhel composes?


Maybe it would be useful, if it is doable, to add a short description of the test at the beginning of the ks.in file (we do it for some of the tests) from which - reading it for both tests - would be clear what is specific for each of the test. (Way of referring to raid from volgroup command?)

Other than that looks good to me, thank you!

vojtechtrefny:
> A typo on the commit message: "... and also make sure the preferred name is not changed by or pykickstart."
> 

Fixed.

> How should we deal with failures on rhel-9 and rhel-10 (/test-os-variants)?
> 
>     * disable them with skip-on-rhel
> 
>     * disable them with a reference to an jira or github issue and update the disabled tests table?, and reenable when they hit rhel composes?
> 

I always forget the same tests are running both on Fedora and RHEL. The pykickstart bug is planned to be fixed in both RHEL 9 and 10 so I added these two test cases to skip-testtypes with the Jira issues referenced.

> 
> Maybe it would be useful, if it is doable, to add a short description of the test at the beginning of the ks.in file (we do it for some of the tests) from which - reading it for both tests - would be clear what is specific for each of the test. (Way of referring to raid from volgroup command?)
> 

Done.

rvykydal:
> /test-os-variants

https://github.com/rhinstaller/kickstart-tests/actions/runs/14614842369/job/41000628075
Unfortunately, all raid tests are broken on rawhide / daily-iso currently: https://github.com/rhinstaller/kickstart-tests/issues/1414

rvykydal (approved):
Looks great to me, thank you.

vojtechtrefny:
Rebased to the latest main.

PR #1375, opened 2025-02-26
Create new testtype dns

rvykydal:
We will be adding more tests for name resolution for eDNS and for image mode.


jstodola (approved):
Thanks

PR #1381, opened 2025-03-04
Add network from kickstart version of e2e edns test

rvykydal:
In this test only global dns is set via boot options. Everything else comes from kickstart, including default dhcp network.

Related: INSTALLER-4096

### DIFF
--- a/dns-global-exclusive-tls-ksnet.ks.in
+++ b/dns-global-exclusive-tls-ksnet.ks.in
@@ -3,10 +3,15 @@
 # Partial tests of reqiured features are in 'certificate' and 'dns-global-bootopts'
 # tests.
 
-# Use defaults.
+# Use defaults, but no network
+%ksappend common/common.ks
 %ksappend repos/default.ks
-%ksappend common/common_no_payload.ks
+%ksappend l10n/default.ks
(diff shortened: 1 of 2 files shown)


jstodola (commented):
One encrypted DNS server is set by the `rd.net.dns` boot option, another DNS (without encryption) is possibly set by the `network --bootproto=dhcp` kickstart command.
The test checks that hostname resolution works, but it doesn't care what DNS server is really used.
Could you please add a check that the eDNS server is really used for hostname resolution?

rvykydal:
> One encrypted DNS server is set by the `rd.net.dns` boot option, another DNS (without encryption) is possibly set by the `network --bootproto=dhcp` kickstart command. The test checks that hostname resolution works, but it doesn't care what 
DNS server is really used. Could you please add a check that the eDNS server is really used for hostname resolution?

@jstodola I am not sure how to check the use of eDNS server.
I realized that I'll rather add this use case as another test with same setup we have for the 3 already existing and handle dhcp configured networking separately. Added a commit doing it.
And explore how to check the DNS server that was used.

jstodola (approved):
It looks good, thank you!

PR #1386, opened 2025-03-07
Add edns e2e test for known cert

rvykydal:


dns-global-exclusive-tls-2.sh
```diff
@@ -0,0 +1,34 @@
...
+# Red Hat Author(s): Radek Vykydal <rvykydal@redhat.com>
+
+# Ignore unused variable parsed out by tooling scripts as test tags metadata
+# shellcheck disable=SC2034
+TESTTYPE=${TESTTYPE:-"network dns skip-on-rhel gh1380"}
+
+. ${KSTESTDIR}/functions.sh
+
+kernel_args() {
+    echo ${DEFAULT_BOOTOPTS} rd.net.dns=dns+tls://1.1.1.1 rd.net.dns-resolve-mode=exclusive rd.net.dns-backend=dnsconfd ip=10.0.2.200::10.0.2.2:255.255.255.0:::none
```
jstodola, line 27: The "1.1.1.1" certificate is issued also for the `1.1.1.1` IP address, so this argument works without providing the host name `one.one.one.one`. But, for a better test coverage, I suggest changing one of the kernel arguments in this PR to also include the DNS host name: `rd.net.dns=dns+tls://1.1.1.1#one.one.one.one`.

rvykydal: Good idea.

PR #1394, opened 2025-03-12
Check that allow-discards flag is set by default on LUKS2 devices

vojtechtrefny:
We'll need to wait for `python-blivet-3.12.0-2.fc43` to make it through rawhide build magic for this to work (it should be available in tomorrows compose).

Depends on https://github.com/storaged-project/blivet/pull/1347 and https://github.com/storaged-project/libblockdev/pull/1095

### DIFF
--- a/autopart-luks-2.ks.in
+++ b/autopart-luks-2.ks.in
@@ -36,6 +36,14 @@ if [[ "$result" != "2" ]] ; then
     exit 1
 fi
 
+# Check if the 'allow-discards' flag is set.
+result=$(cryptsetup luksDump ${crypted} | grep "Flags" | cut -d: -f2 | tr -d "\t\n ")
+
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
The test passes with the latest rawhide ISO `Fedora-Everything-netinst-x86_64-Rawhide-20250313.n.0.iso`

rvykydal:
> /test-os-variants
https://github.com/rhinstaller/kickstart-tests/actions/runs/13854869262

We probably want to disable the test on rhel10 and rhel9, similar as in for example https://github.com/rhinstaller/kickstart-tests/commit/2d0c58d919a9f2b76847d099f6716e13a9ea94c0
I think rather in this PR.

rvykydal (approved):
Looks good to me, thank you.

PR #1407, opened 2025-04-08
DO NOT MERGE: check missing packages do break installation

M4rtinK:


M4rtinK:
Yep, indeed packages named in the %packages section do fail the installation run if not found in the specified repositories. :-) Closing the PR now.

PR #1408, opened 2025-04-08
Fix lsinitrd check for images with kernel version different from repo

rvykydal:
For example images generated on anaocnda PR CI.
Like https://github.com/rhinstaller/anaconda/pull/6338#issuecomment-2786199341


jstodola (approved):
Thanks

PR #1422, opened 2025-05-21
Update ram size for stage2 from net

rvykydal:


functions.sh
```diff
@@ -468,6 +471,7 @@ get_timeout() {
 stage2_from_ks() {
     # Return true to get stage2 location from kickstart (inst.stage2 option will not be supplied).
     # returns: "true" or "false"
+    # If used get_required_ram should be set to STAGE2_FROM_COMPOSE_RAM_SIZE value.
```
jstodola, line 474: I think a comma is missing in the sentence to make it more clear: "If used, get_required_ram should be set..."

jstodola (commented):
Just one small suggestion, otherwise it looks good, thanks.

jkonecny12 (approved):
Looks great, thank you!

PR #1436, opened 2025-06-16
Update VM RAM for tests downloading installer image (gh1420) (gh1389)

rvykydal:
The stage2 image has increased:
last increase (9.7) size: 1024MB
current 9.7 size: 1316MB

It came earlier than I'd expect and with bigger rise than I'd expect. See the last update ~ 1 month ago:
commit d9b783348002e06e7b43c4283d3928892ed5f277


rvykydal:
First check the current state.

PR #1440, opened 2025-06-18
Disable lvm-cache-* on daily-iso temporarily (gh1437)

rvykydal:
The test is failing on daily iso, tracked in gh1437.


jikortus (approved):
Ack :-)

PR #1441, opened 2025-06-19
Do not report failed checks to RESULT after reboot

jkonecny12:
Checks after reboot have to report errors to stdout instead of standard /root/RESULT. These outputs are then processed by the `success_on_first_boot` fragment. Outputs written to /root/RESULT seems to be ignored.


jkonecny12:
RHEL-10 is now completely broken as we are wating for a kernel fix.

PR #1464, opened 2025-08-01
Synchronize classify script issues with the repo (open/closed)

rvykydal:
Assisted by: claude

Created by these prompts:

The scripts/classify-failures script contains two lists of issues. ISSUES list should contain only the repository issues that are open, the CLOSED_ISSUES list should contain olny the repository issues that are closed. Can you check with the repository issues and suggest a patch for the classify-failures script that would reflect current state of the repository issues?

Save the last prompt as a command named classify-sync-with-repo

### DIFF
--- a/.claude/commands/classify-sync-with-repo.md
+++ b/.claude/commands/classify-sync-with-repo.md
@@ -1,0 +1,21 @@
+# classify-sync-with-repo
+
+## Description
+Synchronize the classify-failures script with the current state of repository issues by moving closed issues from ISSUES list to CLOSED_ISSUES list and open issues from CLOSED_ISSUES list to ISSUES list.
+
+## Usage
+```
+classify-sync-with-repo
+```
+
+## What it does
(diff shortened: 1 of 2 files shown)


abadger:
If I understand correctly, the prompt in `.claude/commands` is run by Claude Code to update the lists held in variables in `scripts/classify-failures`?

Due to the copyright status of AI generated artifacts, I would move the lists into either a separate python file or a true data file (yaml, json, toml, csv, whatever) so that we keep a clean separation between the potentially uncopyrightable portions generated by Claude and the code which we retain copyright on. (IANAL but my guess is that since Claude is only moving lines that we wrote around, we probably have a strong case that we retain copyright over that material but it seems fine to separate the data from the code anyways, so it seems like a good step to take now, before things get muddy.)

rvykydal:
@abadger makes a lot of sense to me.

rvykydal:
@abadger I opened a new version of the PR: https://github.com/rhinstaller/kickstart-tests/pull/1483

PR #1480, opened 2025-08-05
Enable network bond pre tests on rhel-like platforms

rvykydal:
The support for the configration has already arrived to rhel.


rvykydal:
> /test-os-variants
https://github.com/rhinstaller/kickstart-tests/actions/runs/16749435526
daily-iso error/missing is irrelevant for this issue.

jstodola (approved):

PR #1483, opened 2025-08-07
Update classify failures with closed issues v2

rvykydal:
The original goal of the PR was to update issue patterns lists to reflect the current state of the repository (open vs closed issues) using Claude AI assistent, but on a [suggestion](https://github.com/rhinstaller/kickstart-tests/pull/1464#issuecomment-3154202929) by @abadger I first separated the data out of the script (first commit) and let Claude work only with the data file (second commit).

### DIFF
--- a/scripts/classify-failures.yml
+++ b/scripts/classify-failures.yml
@@ -1,0 +1,188 @@
+issues:
+#  - description: "<ISSUE DESCRIPTION PREFERABLY WITH GITHUB ISSUE REFERENCE>"
+#    last_matching_re:
+#      - "<REGULAR EXPRESSION FOR GREP TO FIND THE LAST MATCHING LINE>"
+#      - "<REGULAR EXPRESSION FOR PYTHON TO MATCH AGAINST THE FOUND LINE>"
+#
+#  - description: "<ISSUE DESCRIPTION PREFERABLY WITH GITHUB ISSUE REFERENCE>"
+#    first_grep: "<REGULAR EXPRESSION FOR GREP TO BE LOOKED FOR>"
+
+  - description: "[675] ([1]) https://github.com/rhinstaller/kickstart-tests/issues/675"
+    last_matching_re:
+      - "anaconda:"
+      - ".*Configuring \\(running scriptlet for\\): rootfiles.*"
+
+  - description: "[TODO] https://github.com/rhinstaller/kickstart-tests/issues/TODO"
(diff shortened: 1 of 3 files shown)


abadger (changes requested):
Nice work.  The data file looks good.  In the future, we could validate the data file before processing it, then we would be able to depend on the format of the data being correct in the processing code.

I left a few comments on where to handle errors.  But other than that, I think this looks good.

scripts/classify-failures.yml
```diff
@@ -0,0 +1,205 @@
...
+    first_grep: "Payload error.*Failed to download metadata for repo"
+
+
+  - description: "[786] https://github.com/rhinstaller/kickstart-tests/issues/786"
+    first_grep: "Traceback.*Failed to activate swap on /dev/md/test-raid-ddf_0p2: No such file or directory"
+
+
+
+
+
```
abadger, line 36: This is in no way a blocker; I'm just curious.  Did Claude Code decide to just make some of the spaces between entries larger than others?

abadger (approved):
This looks good.  Approved!

scripts/classify-failures.yml
rvykydal: I think there remained gaps (blank lines) after removing items
https://github.com/rhinstaller/kickstart-tests/pull/1483/commits/d3727ebf038f7d19afc2495b0e9fd89d757cbd95
Should have told it to fix that :) I'll do it in a separate PR.

rvykydal:
@abadger I needed to do an update of the PR for failing static test:
https://github.com/rhinstaller/kickstart-tests/actions/runs/16938509926/job/48000982054?pr=1483
In the end I replaced the collections import and check with try/except block but still bumped the pylint version in https://github.com/rhinstaller/kickstart-tests/pull/1490

abadger:
> @abadger thank you for the review. I needed to do an update of the PR for failing static test: https://github.com/rhinstaller/kickstart-tests/actions/runs/16938509926/job/48000982054?pr=1483 In the end I replaced the collections import and check with try/except block but still bumped the pylint version in #1490

Cool.  That change to `try: except:` looks good.  I am guessing the pylint failure could be some sort of python version mismatch (that particular python module has moved around) but I like your new approach better anyways :-)

abadger (approved):
The new changes look good.  I like the `try: except:` approach that you've switched to better than the `isinstance` check.  +1 from me.

PR #1495, opened 2025-09-02
Add check for kernel messages level to a text ui test

rvykydal:
Related: RHEL-110525


jstodola (approved):
It looks good to me, thanks.

PR #1502, opened 2025-09-18
Fix fstab options check of mountpoint-assignment-2 test

rvykydal:
The test used to be passing because the record was not in the fstab at all, which is fixed in
https://github.com/rhinstaller/anaconda/pull/6639. With the fix the check is actually applied so its expected value needs to be fixed.

Also check the record was found in fstab.


rvykydal:
Pending on https://github.com/rhinstaller/anaconda/pull/6639

KKoukiou (approved):
Tbh i dont get how the defaults option appeared

jstodola (approved):

PR #1504, opened 2025-09-19
Update mountpoint assignment

rvykydal:
Followup of https://github.com/rhinstaller/kickstart-tests/pull/1502.
The test is pending on fixes getting in RHEL 10 and RHEL 9 (will probably disable it temporarily)

The 'plain' in the name of the test refers to 'plain' partitioning scheme.


jstodola (approved):
It looks good to me, thanks.

PR #1505, opened 2025-09-22
[INSTALLER-4343] Add test for deprecated RHEL-10 commands

jikortus:
It covers *deprecated* `module` and `vnc` commands and `--teamconfig` and `--teamslaves` options of the `network` command and replaces a downstream test that would need an update anyway.

The *removed* commands aren't checked as, AFAIK, we need the installation to finish, which contradicts with the nature of removed commands that make the installation process stop at the very beginning.

To test deprecation of the `module` command, some modular content is needed - I've packaged it in a tarball in the root similarly to already existing tarballs for `preexisting-btrfs` and `raid-ddf` tests as unfortunately I'm not aware of a better place.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -408,6 +408,14 @@ Service jobs
 These jobs don't have any particular infrastructure requirements. They run on
 GitHub's infrastructure and can be run manually by a developer.
 
+External test data
+------------------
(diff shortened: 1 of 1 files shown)


jstodola (approved):
It looks good to me, thanks.

PR #1506, opened 2025-09-24
infra: fix for broken /test-tmt kickstart-tests repository workflow

adamkankovsky:


rvykydal (approved):
I would agree with merging this so we can try to run the workflow in the reopository. It is broken anyway, and well isolated from other workflows IMO.

PR #1517, opened 2025-09-30
WIP [INSTALLER-4255] Add fedora-eln platform

jikortus:


jikortus:
@rvykydal, can you please have a look at the PR? The /test-os-variants test is failing because PR #1519 needs to get merged first.

fragments/platform/fedora-eln/payload/unified_packages.ks
```diff
@@ -0,0 +1,7 @@
+# CentOS 10 packages section from unified DVD.
```
rvykydal, line 1: 'CentOS 10' -> 'Fedora ELN'

jikortus:
Thank you very much, @rvykydal! I've replaced the RHEL/CentOS strings in the fragments, sorry I left them when I copied over the original fragments. The initial `skip-on-fedora-eln` tags were basically added to tests with `skip-on-centos10` and `skip-on-rhel10`, which I believe should be (mostly) reasonable.

I've also removed the commit adding `fedora-eln` to `/test-os-variants` and created a new pull request + several others in the same manner as you'd done with the `centos10` platform enablement; those will need to be rebased later on when this PR is merged.

BTW, also thank you for reviewing PR #1519, but I can't merge it yet, as it seems that we are having some issues with `/test-tmt` again :(.

jstodola (commented):
There are two commits `Add fedora-eln to script generating launch platform arguments`, which is confusing. Otherwise it looks good to me.

PR #1521, opened 2025-10-02
Add fedora-eln to /test-os-variants GH workflow

jikortus:


rvykydal:
@jikortus I think we want to merge it so that we can test enabling/disabling of the individual tests ?

PR #1522, opened 2025-10-02
Add fedora-eln permian daily test plan

jikortus:
Resolves: INSTALLER-4255


rvykydal (approved):
Looks good to me.
This code actually belongs to the https://github.com/rhinstaller/kickstart-tests/pull/1517/commits/19d99b51b6507837bedf853b0c1aae2a7e8c3a5f (adding to daily tests requires adding template for the platform plan) - it was oversight in my centos-10 patch series and I had to add it later.

PR #1552, opened 2025-11-18
Disable bootc-basic on rhel 10 until the feature is ported

rvykydal:


rvykydal:
> /test-os-variants

Rawhide, fedora-eln, and daily-iso failures are expected. The point of this PR is disabling on rhel 10.

PR #1575, opened 2025-12-09
Disable bootc luks on fedora eln

rvykydal:
Also enable bootc-* tests on fedora (they were passing on disabled tests runs).


rvykydal:
> /test-os-variants

https://github.com/rhinstaller/kickstart-tests/actions/runs/20056794309

The failure on centos10 is something we should fix in another PR - @KKoukiou should we disable the tests on centos variants or are we expecting this to be supported on centos 10?

Issue #604, opened 2021-10-21
packages-weakdeps: "gnupg2 --recommends has changed, test needs to be updated"


rvykydal:
This is just an import of the https://bugzilla.redhat.com/show_bug.cgi?id=1960279 to be tracked here.

Scenario: rhel-9

Kickstart test packages-weakdeps started to fail with:

"gnupg2 --recommends has changed, test needs to be updated"

First seen yesterday in Daily run #182 (https://github.com/rhinstaller/kickstart-tests/actions/runs/837151646 scenario) rhel9

With update from gnupg2-2.2.27-5.el9 to gnupg2-2.3.1-1.el9 pinentry and gnupg2-smime were moved from "Recommends" to "Suggests"

The kickstart test needs to be updated.

PR #1583, opened 2026-01-19
Disable packages-weakdeps on rhel 10 (gh604)

rvykydal:
Disable packages-weakdeps test on rhel 10 because of issue gh604.
see https://github.com/rhinstaller/kickstart-tests/issues/604#issuecomment-3767387057


jstodola:
Shouldn't it be disabled in `rhel10_centos10_disabled_array` instead? Sooner or later, gnupg2-2.4.5-3.el10_1 will be included in CentOS Stream 10 as well.

PR #1591, opened 2026-01-29
Bump containers/runner/Dockerfile to use Fedora 43

KKoukiou:


rvykydal (approved):
Looks good to me.
I guess you tested it locally.

rvykydal (commented):
Do we want to disable core dumps?
It can be valuable debugging info. IIRC we had instances of kickstart test failures (races reproducible sanely only when running in many instances via kickstart test framework) where we inspected the dump with gdb, so being unable to run the test with the dump enabled might hit us in the future?
What do you think @jstodola ?

PR #1598, opened 2026-02-02
Fix fedora-live-image-build.ks.in: remove --iscrypted from locked rootpw

KKoukiou:
The rootpw command was using both --iscrypted and --lock flags together, which is invalid. The --iscrypted flag requires a password hash value, but --lock is used to lock the account without setting a password.

Recent releases from shadow-utils may have made password validation checks tighter, causing 'Unable to set password for new user: status=1' errors when --iscrypted is used without a valid password hash.

Fix by removing --iscrypted and using only --lock.

### DIFF
--- a/fedora-live-image-build.ks.in
+++ b/fedora-live-image-build.ks.in
@@ -14,7 +14,7 @@ xconfig  --startxonboot
 # Keyboard layouts
 keyboard 'us'
 # Root password
-rootpw --iscrypted --lock locked
+rootpw --lock
 # System language
 lang en_US.UTF-8
 # Shutdown after installation


AdamWill:
Note the changes to shadow were broken in [various ways](https://github.com/shadow-maint/shadow/issues/1483).

I'm not totally sure `lock` and `crypted` are fundamentally incompatible. AIUI it's valid, in some sense, to have an account with a valid password that is 'locked' by adding a `!` prefix to it; the account can then be 'unlocked' by removing the prefix, at which point it becomes accessible using the specified password. (shadow-utils may have a command that can do this, I'm not sure offhand).

If we just want to make the account entirely inaccessible, yes, we don't need `crypted`, probably, but the result of that is likely a bare `!`, which shadow 4.19 also did not accept initially.

shadow-utils-4.19.0-5 should already have fixed both cases (valid hashes prefixed with `!`, and bare `!`).

PR #1610, opened 2026-02-19
Force restorecon on 00-keyboard.conf for container iso builds (gh1607)

rvykydal:
If boot.iso is created (by lorax) in a container the
/etc/X11/xorg.conf.d directory created in the image has container_file_t
and therefore /etc/X11/xorg.conf.d/00-keyboard.conf created during
installation will have the same type. This type will require forcing by
-F option when restoring selinux context at the end of installation.
Otherwise it would not be applied with "not reset as customized by
admin" message.

In non-container / official builds the file would have (as all files in
the image) system_u:object_r:var_lib_t:s0 which would be restored at the
end of installation without the need of -F option.

Related: INSTALLER-4605


rvykydal:
> /test-os-variants
https://github.com/rhinstaller/kickstart-tests/actions/runs/22180165860
Test failures are unrelated.

Issue #803, opened 2022-09-27
It is not possible to pass additional argument containing space via launch --run-args option


rvykydal:
We hit the issue when trying to pass boot options via `launch --run-args`: https://github.com/rhinstaller/kickstart-tests/pull/797
To me it seems to require some advanced bash quoting & whitespace mastery. Or perhaps changing the --run-args to --run-arg with multiple values.

PR #797, opened 2022-09-21
Separate values in KSTEST_EXTRA_BOOTOPTS by a semicolon

rvykydal:
So that multiple values can be passed using --run-args launcher option, for example

--run-args '-eKSTEST_EXTRA_BOOTOPTS=inst.xtimeout=100;inst.sshd=1'

PR #1647, opened 2026-04-14
launch: add --run-arg option to support arguments with spaces (gh#803)

adamkankovsky:
The existing --run-args option word-splits its value, making it impossible to pass container runtime arguments whose values contain spaces (e.g. --env=KSTEST_EXTRA_BOOTOPTS="inst.text inst.sshd").

Add a new --run-arg (singular) option that preserves quoting: each --run-arg value is appended as a single element to a bash array, so spaces in values are not lost. The option can be repeated.

The legacy --run-args is kept for backward compatibility; it still word-splits, but now also accumulates into the same array, so it can be specified multiple times as well.

issue: [803](https://github.com/rhinstaller/kickstart-tests/issues/803)

### DIFF
--- a/containers/runner/README.md
+++ b/containers/runner/README.md
@@ -147,8 +147,16 @@ the graphical console for a virtual machine with a command:
 
     virt-viewer -c qemu+tcp://<IP>/session
 
-To change of the boot options for the tests (for example to add `inst.text`), please use
-`--run-args="--env KSTEST_EXTRA_BOOTOPTS=inst.text"` parameter for the `launch` script.
+To change the boot options for the tests (for example to add `inst.text`), please use
+the `--run-arg` parameter for the `launch` script:
+
+    containers/runner/launch --run-arg="--env=KSTEST_EXTRA_BOOTOPTS=inst.text" keyboard
+
+To pass values containing spaces, use `--run-arg` (can be repeated):
+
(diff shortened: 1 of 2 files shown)


containers/runner/README.md
```diff
@@ -147,8 +147,16 @@ the graphical console for a virtual machine with a command:
...
-To change of the boot options for the tests (for example to add `inst.text`), please use
-`--run-args="--env KSTEST_EXTRA_BOOTOPTS=inst.text"` parameter for the `launch` script.
+To change the boot options for the tests (for example to add `inst.text`), please use
+the `--run-arg` parameter for the `launch` script:
+
+    containers/runner/launch --run-arg="--env=KSTEST_EXTRA_BOOTOPTS=inst.text" keyboard
+
+To pass values containing spaces, use `--run-arg` (can be repeated):
+
+    containers/runner/launch --run-arg="--env=KSTEST_EXTRA_BOOTOPTS=inst.text inst.sshd" keyboard
```
rvykydal: Does this really work?

In the end we are using ';' to separate the options (https://github.com/rhinstaller/kickstart-tests/pull/797). I think it is actually better then using spaces. Also, just reverting the PR to use spaces does not work for me, it has some issues.

rvykydal: Maybe we just need another example (or no change at the doc at all)

rvykydal (approved):
Looks good to me, thank you.

PR #1661, opened 2026-05-28
Add tests for certificate --type=anchor

rvykydal:
Resolves: INSTALLER-4752


jstodola (commented):
It looks good to me, however is there a reason to have two separate tests? IMO, they can be merged into one to save resources.

rvykydal:
> It looks good to me, however is there a reason to have two separate tests? IMO, they can be merged into one to save resources.

`certificate-anchor` is independent of any prepared infrastructure while the `certificate-anchor-curl` depends on the server to download from - currently static rtt1 server. So if the latter one (curl) stops working, we can disable it easily while still testing the feature quite well.

Other solution could be merge the tests and disable (comment out) the curl testing in the test case of infra issues.

jstodola:
OK, it makes sense. Thanks for the tests!

jstodola (approved):

rvykydal:
The test in CI passed: https://github.com/rhinstaller/anaconda/pull/7084#issuecomment-4573471494.
Let's wait a bit before merging for the feature in daily-iso.

rvykydal:
Merging, daily-iso passes, rawhide will pass after the next build.
https://github.com/rhinstaller/kickstart-tests/actions/runs/26638881574

PR #1663, opened 2026-06-05
Remove almost duplicate ntp tests

jstodola:
There are the following tests executed in both graphical and text UI:
* ntp-nontp-without-chrony / ntp-nontp-without-chrony-gui
* ntp-with-nontp / ntp-with-nontp-gui
* ntp-without-chrony / ntp-without-chrony-gui

To save resources, pick only one UI variant for each test and remove the other one.


rvykydal (approved):
Looks good to me, thank you.

PR #1664, opened 2026-06-05
Ensure firewalld state matches systemd preset

jstodola:
The previous check was incorrect (checking a non-existing firewall service, insead of the firewalld service).
Removed the previous check for firewall enabling and added a validation to ensure that the firewalld service state aligns with its systemd preset.

Generated-by: Cursor


rvykydal (approved):
Looks good to me, thank you.

PR #1673, opened 2026-06-12
Fix cleanup of /tmp/lorax-images before boot.iso build

KKoukiou:
The rm -rf was introduced in 98b3e58f when extracting the reusable workflow (because now we might have leftovers in that directory from WebUI or non WebUI job and we don't want to mix these artifacts) but ran without sudo.


rvykydal (approved):

KKoukiou:
not waiting for the tests. it's not affecting those anyway.

PR #1675, opened 2026-06-15
infra: Add daily-iso-webui Permian scenario for WebUI boot.iso

KKoukiou:
Run the container test against webui-boot.iso from the images-webui artifact and teach fetch_daily_iso.sh to select artifact and ISO name.

Assisted-by: Cursor (Composer)


rvykydal (approved):
Looks good to me

KKoukiou:
Tried locally and the test passes.

PR #1676, opened 2026-06-16
infra: Add daily-iso-webui to test-os-variants workflow

KKoukiou:
Run PR /test-os-variants jobs against the daily WebUI boot.iso. For the daily-iso-webui variant, skip all tests except lang via skip-testtypes and generate the daily Permian test plan from a template, similar to `testlib/test_plans/daily-daily-iso.plan.yaml.j2` plan.

Assisted-by: Cursor (Auto)

### DIFF
--- a/.github/workflows/test-os-variants.yml
+++ b/.github/workflows/test-os-variants.yml
@@ -117,7 +117,7 @@ jobs:
       GITHUB_TOKEN: /home/github/github-token
     strategy:
       matrix:
-        os-variant: [daily-iso, rawhide, fedora-eln, rhel9, rhel10, centos10]
+        os-variant: [daily-iso, daily-iso-webui, rawhide, fedora-eln, rhel9, rhel10, centos10]
       fail-fast: false
 
     steps:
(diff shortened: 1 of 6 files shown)


rvykydal (commented):
Maybe you have a further plan (something thought over I don't see) with using test categories to skip tests. I think you'll need better granularity anyway, but it can be next step.

As an alternative it could be worth to use a new testtype `webui` which would be added in steps, and the test plan would check for this type. But I guess the approach needs more thinking and can hit some drawback.
Or maybe even using `skip-on-webui` added to all tests as a baseline.

Anyway I don't mind starting with this PR, it should not break current workflows.

KKoukiou:
@rvykydal I also thought about adding a new webui testtype. but generally we want 'mostly' all tests to run ith webui as well - soonish. Only the ui tests are irrelevant I guess? Maybe some more. But it's nicer to have a long list and try to minimize it, than adding to more and more tests the webui type. 

Unless we agree that we don't want to run webui against all relevant tests for now, since it's not yet a deliverable.

rvykydal (approved):
Looks good to me.

PR #1681, opened 2026-06-22
infra: Enable smoke tests for daily-iso-webui scenario

KKoukiou:
Unblock smoke test categories on the WebUI daily ISO and skip TUI smoke tests that do not apply via skip-on-daily-iso-webui.

### DIFF
--- a/functions.sh
+++ b/functions.sh
@@ -479,5 +479,9 @@ stage2_from_ks() {
 
 # RAM size of VM for the test in MiB
 get_required_ram() {
-    echo "2048"
+    if [ "${KSTEST_WEBUI}" = "true" ]; then
+        echo "2560"
+    else
+        echo "2048"
+    fi
 }
--- a/post-nochroot-lib-network.sh
+++ b/post-nochroot-lib-network.sh
@@ -23,6 +23,12 @@ function check_bridge_has_slave_nochroot() {
     fi
 }
 
+# anaconda_uses_webui
+# Return success if anaconda ran the WebUI instead of the GTK UI.
+function anaconda_uses_webui() {
+    grep -qE 'cockpit web view has been started' /tmp/anaconda.log
+}
+
 # check_gui_configurations
 # Works only for gui installations.
 # Checks that the connections configurable in Network Spoke are those corresponding to configuration files of devices
@@ -35,6 +41,10 @@ function check_gui_configurations() {
     # Pass the test if not in graphical mode
     grep -q "Display mode is set to.* graphical mode" /tmp/anaconda.log
     if [[ $? -ne 0 ]]; then
+        return
+    fi
+
+    if anaconda_uses_webui; then
         return
     fi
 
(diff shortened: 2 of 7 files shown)


rvykydal:
> /test-os-variants --testtype smoke

I see some failures in tests, not looked in detail though...
https://github.com/rhinstaller/kickstart-tests/actions/runs/27956057340/job/82724660304
Also interestingly reboot-initial-setup-tui seems to be passing so should it be disabled?

KKoukiou:
> > /test-os-variants --testtype smoke
> 
> I see some failures in tests, not looked in detail though... https://github.com/rhinstaller/kickstart-tests/actions/runs/27956057340/job/82724660304 Also interestingly reboot-initial-setup-tui seems to be passing so should it be disabled?

thanks I will check. They are for sure real - example the raid issue reproduced locally. 
I thought that plain rn of test-os-variants https://github.com/rhinstaller/kickstart-tests/pull/1681#issuecomment-4767362759 triggers smoke tests by default

rvykydal:
> ... I thought that plain rn of test-os-variants [#1681 (comment)](https://github.com/rhinstaller/kickstart-tests/pull/1681#issuecomment-4767362759) triggers smoke tests by default

It tries to detect the changed tests, and if it does not detect any, it runs the smoke tests.
In this PR ui_text_interactive.sh is modified so only this test is run.

KKoukiou:
@M4rtinK for when you rerview: the failures are expected. Or failing tests are being skipped

post-nochroot-lib-network.sh
```diff
@@ -23,6 +23,12 @@ function check_bridge_has_slave_nochroot() {
     fi
 }
 
+# anaconda_uses_webui
```
M4rtinK, line 26: Just a note related to the commit message:

"WebUI does not log GTK Network spoke device configurations, so
check_gui_configurations does not work on WebUI boot.iso runs."

I wonder if perhaps there is still some logic in the GTK/GUI code that should be moved to the backend & the moved logic can then log the data tests need ?

M4rtinK (approved):
Looks good to me. :)

post-nochroot-lib-network.sh
KKoukiou: Indeed we should not be checking front end logs...

functions.sh
```diff
@@ -479,5 +479,10 @@ stage2_from_ks() {
 
 # RAM size of VM for the test in MiB
 get_required_ram() {
-    echo "2048"
+    # Webui tests need more memory, check for environment variable
+    if [ -n "${KSTEST_WEBUI_RAM}" ]; then
+        echo "${KSTEST_WEBUI_RAM}"
+    else
+        echo "2560"
```
rvykydal, line 483: Shouldn't this be 2048?

rvykydal (commented):
One question.

rvykydal (approved):

functions.sh
KKoukiou: Fixed

PR #1685, opened 2026-07-01
services: test enabling a socket-activated unit via kickstart

KKoukiou:
Enable rpcbind.socket in the services test and verify it is enabled after installation.

Related: RHEL-178289


jstodola (approved):
It looks good to me, thanks!

PR #1686, opened 2026-07-06
Add bootc UEFI kickstart test without separate /boot partition

KKoukiou:
Test bootc installation with reqpart and a single root partition on UEFI, where /boot lives on / and only /boot/efi is mounted separately.

Related: RHEL-187037


fragments/shared/storage/reqpart_root.ks
jstodola, line 1: I suggest renaming the file to `reqpart_root.ks` only, since the commands are not really related to bootc/ostreecontainer and could be possible reused by other tests as well. But it could be done as a followup PR, together with renaming https://github.com/rhinstaller/kickstart-tests/blob/main/fragments/shared/storage/ostreecontainer_autopart.ks and https://github.com/rhinstaller/kickstart-tests/blob/main/fragments/shared/storage/ostreecontainer_autopart_encrypted.ks in a similar way.

jstodola (approved):
Thanks

M4rtinK (approved):
Agreed on the better naming with Honza, otherwise looks good. :)

PR #1694, opened 2026-07-15
doc: update with a hint for running all enabled tests

rvykydal:


KKoukiou (approved):
We can just add a --os-variant KS test parameter to avoid opening dummy anaconda PRs.

rvykydal:
> We can just add a --os-variant KS test parameter to avoid opening dummy anaconda PRs.

Yes, actually this may be pretty low-hanging fruit.
And also allow to set if we want to run disabled tests or not.

PR #1698, opened 2026-07-20
Migrate snapshot-pre test to UEFI.

rvykydal:
Resolves: INSTALLER-4820

It seems to be working well with UEFI

https://redhat.atlassian.net/browse/INSTALLER-4820


jikortus (approved):
Thank you.

PR #1702, opened 2026-07-21
Add os variants support to test os variants

rvykydal:
https://redhat.atlassian.net/browse/INSTALLER-4827

### DIFF
--- a/.github/workflows/test-os-variants.yml
+++ b/.github/workflows/test-os-variants.yml
@@ -114,8 +114,8 @@ jobs:
             done
             OS_VARIANTS="${OS_VARIANTS%,}]"
             if [ "$OS_VARIANTS" == "[]" ]; then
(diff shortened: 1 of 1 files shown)


jikortus (approved):
This looks very useful, thank you!

KKoukiou (approved):
As a next step we could even extract the logic to a reusable workflow (similar to what we have in anaconda repository - see .github/workflows/define-matrix.yml) which can be reused in other places - for example in scenarios-permian.yml.

rvykydal:
> As a next step we could even extract the logic to a reusable workflow (similar to what we have in anaconda repository - see .github/workflows/define-matrix.yml) which can be reused in other places - for example in scenarios-permian.yml.

@KKoukiou exactly, it occured to me as well. I think we might want to do it when we touch this area next time (also maybe making sure the logic is testable better - I mean better than logic included in a workflow triggered by comment).

PR #1707, opened 2026-07-22
Reduce TEST_JOBS for daily-iso-webui from 16 to 12

rvykydal:
The webui boot.iso requires 25% more RAM per test (commit d73dcd55f6d839170607acfc979b9c4901415b6b) so reduce number of parallel jobs from 16 to 12.


rvykydal:
@KKoukiou I think quite a lot of failures of daily run of daily-iso-webui may be timeouts caused by low resources. Let's try this to see if it so for one daily?
(Also to see which tests are really broken better.)

KKoukiou (approved):

jikortus (approved):

PR #1721, opened 2026-07-24
Disable most of the luks storage tests on webui for gh1715

rvykydal:


rvykydal:
/test-os-variants --os-variants daily-iso-webui --testtype gh1715 --force

KKoukiou (approved):
There are some tests that succeeded but I geuss since it's  a race they might fail on next run

rvykydal:
> There are some tests that succeeded but I geuss since it's a race they might fail on next run

Yes, some have smaller chance to fail, also some might be even missing. We can update later.

PR #1722, opened 2026-07-24
Add --test-copr option to test-os-variants workflow

KKoukiou:
Allow testing with custom packages from COPR repos by building a boot.iso with additional COPRs enabled. Multiple COPRs can be specified comma-separated. Usage:
/test-os-variants --test-copr=owner/project --os-variants=daily-iso /test-os-variants --test-copr=owner1/p1,owner2/p2 --os-variants=daily-iso-webui

Requires --os-variants with exactly one of daily-iso or daily-iso-webui. The webui flag is set automatically based on the selected os-variant.

The reusable ISO build workflow gains a test_copr input and handles COPR URL construction. test-os-variants adds a conditional build-iso job that calls it before running the test matrix.

Assisted-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/daily-boot-iso-rawhide-reusable.yml
+++ b/.github/workflows/daily-boot-iso-rawhide-reusable.yml
@@ -9,6 +9,11 @@ on:
         required: false
         type: boolean
         default: false
+      test_copr:
+        description: COPR repo name(s) to include in the ISO build (comma-separated, e.g. owner/project or @group/project)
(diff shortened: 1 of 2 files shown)


jikortus (approved):
Having a possibility to use content from COPR looks really useful, thank you!

rvykydal (approved):
Looks good to me.

PR #1735, opened 2026-08-21
Replace unsupported msdos label setups

rvykydal:


KKoukiou (approved):
If you want to keep an msdos test just test it with bios.

rvykydal:
> If you want to keep an msdos test just test it with bios.

We keep msdos in other clearpart tests where there are actually cleared by their variants clearpart command, so I thought it is good to have some of the clearpart tests with gpt.

jstodola (approved):

PR #1745, opened 2026-09-02
infra: Add TOCTOU push detection to test-os-variants workflow

rvykydal:
Use the Check Suites API to verify the PR head commit was pushed before the /test-os-variants trigger comment. Enabled by default, can be skipped with --skip-timestamps-check.

The same patch has been already applied to anaconda repository workflow https://github.com/rhinstaller/anaconda/blob/main/.github/workflows/kickstart-tests.yml

### DIFF
--- a/.github/workflows/test-os-variants.yml
+++ b/.github/workflows/test-os-variants.yml
@@ -42,6 +42,12 @@
 # ... run all the tests of type network only on rhel10
 #
 #
+# By default the workflow verifies that the PR head commit was not pushed
+# after the trigger comment (TOCTOU protection). To skip this check:
+#
+#  /test-os-variants --skip-timestamps-check
+#
+#
 # Lastly there is a dry-run mode, that just indicates which tests would be run
 # (useful to check test grouping and overall infra health):
 #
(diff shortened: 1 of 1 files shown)


.github/workflows/test-os-variants.yml
```diff
@@ -126,13 +132,64 @@ jobs:
...
+          if [ "$EARLIEST_CHECK_SUITE" = "null" ] || [ -z "$EARLIEST_CHECK_SUITE" ]; then
+            echo "::error::No check suites found for commit $EXPECTED_SHA."
+            exit 1
+          fi
+
+          # ISO 8601 timestamps - lexicographic comparison works correctly.
+          # Require check suite to be strictly before the comment: the commit
+          # must have been pushed before the maintainer typed the comment,
+          # so its check suite timestamp must be earlier.
+          if [[ ! "$EARLIEST_CHECK_SUITE" < "$COMMENT_TIME" ]]; then
```
jikortus, line 183:
```suggestion
          if [[ "$EARLIEST_CHECK_SUITE" >= "$COMMENT_TIME" ]]; then
```
Could we perhaps simplify the statement by removing the negation? I somehow managed to overlook the exclamation at first and scratched my head about how this is supposed to work :-) (btw not sure if we couldn't even use `>` instead of `>=` to make it 100% bulletproof, even though this may be a rather theoretical concern).

rvykydal: The granularity of the check (timestamp) is 1s so there is < on purpose to be rather on safer side (rather refuse than allow). Creating a comment by user takes almost certainly more than 1s so we should be ok here.

Also I'd rather prefer to keep the code same as in the anaconda repository (in case of potential follow-ups so that the patch is applicable to both repositories).

jikortus: OK, fair enough :+1:.

jikortus (approved):

rvykydal:
Tested by https://github.com/rhinstaller/kickstart-tests/pull/1748

PR #1752, opened 2026-09-15
Add rpm-ostree test without separate /boot partition (INSTALLER-4393)

adamkankovsky:
Add bootc-biosboot-reqpart-root test to verify the INSTALLER-4077 fix: bootupctl must receive the correct /boot path even when /boot is not a separate partition (reqpart + root only layout on BIOS).

Validates that /boot/bootupd-state.json and /boot/loader/entries exist after installation. Uses %post --nochroot with shutdown (no second boot) to avoid BIOS boot failure after virt-install gets --noreboot.


rpm-ostree-container-reqpart-root.sh
```diff
@@ -0,0 +1,26 @@
...
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+# Ignore unused variable parsed out by tooling scripts as test tags metadata
+# shellcheck disable=SC2034
```
KKoukiou, line 23: gh1574 and bootc are unrelated to this. I qwould put TESTYPE bellow the include so that we override it.

KKoukiou:
/test-os-variants --test-copr=packit/rhinstaller-anaconda-webui-1474 --os-variants=daily-iso-webui

KKoukiou:
/test-os-variants --test-copr=packit/rhinstaller-anaconda-webui-1474 --os-variants=daily-iso-webui

KKoukiou (approved):

rpm-ostree-container-reqpart-root.sh
```diff
@@ -0,0 +1,24 @@
...
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+. ${KSTESTDIR}/bootc-uefi-reqpart-root.sh
+
+# Ignore unused variable parsed out by tooling scripts as test tags metadata
+# shellcheck disable=SC2034
+TESTTYPE="skip-on-rhel-9 payload ostree uefi reboot storage"
```
rvykydal: For consistency (and perhaps other reason, don't remember), we usually put the TESTTYPE first and in the included file we check if the TESTTYPE exists, so in bootc-efi-reqpart-root.sh it would be:
```
TESTTYPE=${TESTTYPE:-"skip-on-rhel-9 payload uefi bootc reboot storage gh1574"}
```

KKoukiou: Yes but gh1574 and bootc are unrelated to this. We should not be extending TESTTYPE here I think. https://github.com/rhinstaller/kickstart-tests/pull/1752#discussion_r4014476516

rvykydal: Ah, I got it, you are right.

PR #1756, opened 2026-09-18
Skip default Anaconda-webui COPR when --test-copr is used

adamkankovsky:
The webui-boot.iso build always added the @rhinstaller/Anaconda-webui COPR alongside any --test-copr repo. As dnf selects the best EVR candidate, the usually newer main build silently shadowed the PR package under test, so /test-os-variants --test-copr for an anaconda-webui PR actually tested main.

Only add the default COPR when no custom test COPR is requested.


KKoukiou (changes requested):
This --test-copr is not only relevant for WebUI. Also PR builds should be when rebased always newer compared to main builds.

